PR #187, opened 2026-03-18
feat: add Reasoning Verification extension for MPP

ThoughtProof:
## What

Proposes a "reasoning-verification" extension for MPP that adds a pre-payment verification step — an independent service evaluates the agent's decision logic before payment credentials are issued.

## Why

Identity checks answer "who is this agent?" Payment authorization answers "can this agent pay?" Neither answers "should this agent pay for this, given its reasoning?"

An authorized agent with valid identity can still make poorly-reasoned payments. This extension catches those defects before settlement.

## How

- Server includes `reasoning-verification="required"` in the 402 challenge
- Agent submits decision logic to a verification endpoint
- Verifier returns ALLOW/HOLD/UNCERTAIN/DISSENT with confidence score
- Agent includes verification result in payment credential

Stake-proportional: low-value payments skip verification, high-value payments require multi-model adversarial critique.

## Design Choices

- Provider-agnostic: any verification service can implement the endpoint
- OPTIONAL extension: servers opt in per-resource or per-amount threshold
- Fail-closed for "required", fail-open for "recommended"
- Compatible with x402 and ERC-8183 verification patterns

## Related

- ERC-8004 (identity) + Maiat (behavioral trust) + this extension (reasoning quality) = complete trust stack
- x402 v2 Extensions system has a similar design space

### DIFF
--- a/specs/extensions/draft-payment-reasoning-verification-00.md
+++ b/specs/extensions/draft-payment-reasoning-verification-00.md
@@ -1,0 +1,344 @@
+---
+title: Reasoning Verification Extension for HTTP Payment Authentication
+abbrev: Reasoning Verification
+docname: draft-payment-reasoning-verification-00
+version: 00
+category: info
+ipr: trust200902
+submissiontype: IETF
+consensus: true
+
+author:
+  - name: Raul Jäger
+    ins: R. Jäger
+    email: raul@thoughtproof.ai
+    org: ThoughtProof
+---
+
+## Abstract
+
+This document defines the "reasoning-verification" extension for the
+Payment HTTP Authentication Scheme [I-D.httpauth-payment]. It specifies
+how a server MAY require adversarial multi-model verification of the
+agent's decision logic before accepting a payment credential — ensuring
+that the reasoning behind a payment is sound, not just that the payment
+is authorized.
+
+## Status of This Memo
+
+This Internet-Draft is submitted in full conformance with the provisions
+of BCP 78 and BCP 79.
+
+## Copyright Notice
+
+Copyright (c) 2026 IETF Trust and the persons identified as the document
+authors. All rights reserved.
+
+## Table of Contents
+
+1. [Introduction](#1-introduction)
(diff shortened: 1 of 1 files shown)


kevin-biot:
The reasoning verifier is an LLM evaluating another LLM's output. That is not deterministic external policy ....  it is the LLM-as-policy-judge antipattern with a confidence score attached. Non-deterministic, prompt-injectable, produces no verifiable record of why it decided, and introduces a third-party dependency into every payment flow.

The correct answer to "should this agent pay for this" is a deterministic policy engine enforcing pre-committed scope;  not a second model guessing at the first model's intent.

If the agent's authorised scope is pre-committed in a credential before the transaction, the question answers itself: does the requested payment fall within the declared action class and intent namespace? That is a deterministic check. No LLM required. No $0.01 toll per transaction. No third-party availability dependency.

See §1.9 of the OBO RFC for the full argument: https://github.com/kevin-biot/obo-standard

ThoughtProof:
Fair points on deterministic scope checks — pre-committed credentials that answer "is this payment within the declared action class?" are valuable and complementary to what we're doing.

Two clarifications:

1. The reasoning verification does produce a verifiable record — every verdict is an on-chain attestation with the full reasoning trace. That's the point: not just "approved/denied" but a durable, auditable record of why.

2. Deterministic policy and reasoning verification solve different problems at different layers. Policy engines handle "is this within scope?" — that's a solved problem. But "is the agent's reasoning for this action sound?" is a different question entirely. When agents operate autonomously in complex environments, scope compliance alone doesn't tell you whether the decision makes sense.

Both layers can coexist. Pre-committed scope as the fast first gate, reasoning verification as the deeper check when the stakes warrant it.

Will check out the OBO RFC — always interested in how these approaches can work together.

brendanjryan:
Thank you! We are not interested in enshrining policies on top of MPP right now, and think that this is better done out of protocol. This does not block this from being used with MPP services, just inclusion in the core spec. 

I will consider this proposal if/when we want to consider this domain

PR #213, opened 2026-04-01
feat: add unified EVM charge payment method

0xBreadguy:
## Summary

Adds a unified `evm` payment method for the charge intent, replacing the previously proposed chain-specific methods (`megaeth` in #205, `sei` in #206). A single method works across all EVM-compatible chains — the only differentiator is `chainId` in `methodDetails`.

Follows direction from @gakonst in #205 and #206 to consolidate into a general EVM spec.

## Credential Types

Four credential types, with a clear preference hierarchy:

1. **`type="permit2"` (RECOMMENDED)** — Client signs off-chain EIP-712 Permit2 authorization, server submits on-chain. Preferred because:
   - Server naturally sponsors gas
   - Splits are atomic via `permitBatchWitnessTransferFrom()`
   - Cryptographic challenge binding via `PaymentWitness` witness data
   - No nonce management burden on client

2. **`type="authorization"` (opt-in for EIP-3009 tokens)** — Client signs off-chain `transferWithAuthorization` message, server submits directly to the token contract. For tokens that natively implement EIP-3009 (USDC, EURC). Zero setup — no Permit2 approval needed. Challenge binding via nonce = `challengeHash`, enforced on-chain by the token contract. No splits.

3. **`type="transaction"` (compatible fallback)** — Client signs a standard ERC-20 `transfer` transaction, server broadcasts. For chains without Permit2 or clients that prefer direct signing. Client pays gas. No splits.

4. **`type="hash"` (constrained client fallback)** — Client broadcasts transaction themselves, presents tx hash. For custodial wallets / hardware signers that can't hand off unbroadcast transactions. Client pays gas. No splits. Weaker challenge binding.

## Key Design Decisions

- **Permit2 batch transfers for splits**: All splits execute atomically in a single `permitBatchWitnessTransferFrom()` call. No partial execution risk. Splits require `type="permit2"`.
- **`permitted` is always an array**: Even for single transfers (length 1). No polymorphic fields — cleaner for typed language implementations.
- **Challenge binding via witness / nonce**: Permit2 uses `PaymentWitness { bytes32 challengeHash }` in EIP-712 witness data. EIP-3009 uses the same `challengeHash` as the on-chain nonce — enforced by the token contract itself.
- **Gas sponsorship is structural**: Off-chain signature types (permit2, authorization) = server always pays gas. On-chain transaction types (transaction, hash) = client always pays. No `feePayer` field needed.
- **No chain-specific appendix**: Chain IDs reference [chainlist.org](https://chainlist.org). Implementation details belong in chain documentation, not the protocol spec.
- **Confirmation depth is server policy**: The spec requires a successful tx receipt (1 block inclusion) and leaves finality depth to server discretion.
- **Native token transfers out of scope**: ERC-20 only.
- **Splits require `type="permit2"`**: Authorization, transaction and hash credentials cannot carry splits.

## New file

- `specs/methods/evm/draft-evm-charge-00.md`

## AI Disclosure

This spec was drafted with AI assistance (Claude). All content has been reviewed for technical accuracy, RFC compliance, and alignment with STYLE.md / CONTRIBUTING.md.

### DIFF
--- a/specs/methods/evm/draft-evm-charge-00.md
+++ b/specs/methods/evm/draft-evm-charge-00.md
@@ -291,7 +291,7 @@ Valid values: `"permit2"`, `"authorization"`, `"transaction"`,
 If omitted, servers MUST accept `"transaction"` and SHOULD
 accept `"hash"`. Servers that support Permit2 SHOULD include
 `"permit2"` as the first entry to indicate preference.
-Servers SHOULD only include `"authorization"` when the
+Servers MUST only include `"authorization"` when the
 `currency` token is known to implement EIP-3009.
 Clients SHOULD use the first type in the list that they
 support.
--- a/specs/methods/evm/draft-evm-charge-00.md
+++ b/specs/methods/evm/draft-evm-charge-00.md
@@ -17,7 +17,7 @@ author:
     ins: C. Swenberg
     email: conner.swenberg@coinbase.com
     organization: Coinbase
-  - name: K. Scott
+  - name: Kyle Scott
     ins: K. Scott
     email: kscott@monad.foundation
     organization: Monad Foundation
--- a/specs/methods/evm/draft-evm-charge-00.md
+++ b/specs/methods/evm/draft-evm-charge-00.md
@@ -145,7 +145,7 @@ This specification defines two credential types:
   - No nonce management burden on the client
   - `externalId` is cryptographically bound via witness data
 
-- **`type="authorization"` (opt-in)**: The client signs an
+- **`type="authorization"`**: The client signs an
   off-chain EIP-3009 {{EIP-3009}} `transferWithAuthorization`
   message. The server submits it to the token contract. This
   is available for tokens that natively implement EIP-3009
--- a/specs/methods/evm/draft-evm-charge-00.md
+++ b/specs/methods/evm/draft-evm-charge-00.md
@@ -53,6 +53,12 @@ informative:
   ERC-20:
     title: "Token Standard"
     target: https://eips.ethereum.org/EIPS/eip-20
+  EIP-3009:
+    title: "Transfer With Authorization"
+    target: https://eips.ethereum.org/EIPS/eip-3009
+    author:
+      - name: Peter Jihoon Kim
+    date: 2019-09
   PERMIT2:
     title: "Permit2"
     target: https://github.com/Uniswap/permit2
@@ -68,9 +74,12 @@ method in the Payment HTTP Authentication Scheme
 exchange one-time ERC-20 token transfers on any EVM-compatible
 blockchain.
 
-Three credential types are supported: `type="permit2"`
+Four credential types are supported: `type="permit2"`
 (RECOMMENDED), where the client signs an off-chain Permit2
 authorization and the server submits the transfer;
+`type="authorization"` (opt-in for EIP-3009 tokens), where
+the client signs an off-chain transfer authorization and
+the server submits it directly to the token contract;
 `type="transaction"`, where the client signs and the server
 broadcasts a standard ERC-20 transfer transaction; and
 `type="hash"` (optional fallback), where the client
@@ -135,6 +144,17 @@ This specification defines two credential types:
   - Split payments are atomic via batch transfers
   - No nonce management burden on the client
   - `externalId` is cryptographically bound via witness data
+
+- **`type="authorization"` (opt-in)**: The client signs an
+  off-chain EIP-3009 {{EIP-3009}} `transferWithAuthorization`
+  message. The server submits it to the token contract. This
+  is available for tokens that natively implement EIP-3009
+  (e.g., USDC, EURC). Benefits:
+
+  - No Permit2 approval prerequisite — zero setup
+  - Server naturally sponsors gas
+  - Challenge binding via the on-chain nonce
+  - Simpler signature structure
 
 - **`type="transaction"`**: The client signs a complete ERC-20
   `transfer` transaction. The server broadcasts it. This is the
@@ -265,11 +285,14 @@ any particular set of chains.
 Servers MAY indicate accepted credential types via the
 `credentialTypes` field in `methodDetails`:
 
-Valid values: `"permit2"`, `"transaction"`, `"hash"`.
+Valid values: `"permit2"`, `"authorization"`, `"transaction"`,
+`"hash"`.
 
 If omitted, servers MUST accept `"transaction"` and SHOULD
 accept `"hash"`. Servers that support Permit2 SHOULD include
 `"permit2"` as the first entry to indicate preference.
+Servers SHOULD only include `"authorization"` when the
+`currency` token is known to implement EIP-3009.
 Clients SHOULD use the first type in the list that they
 support.
 
@@ -283,7 +306,8 @@ Splits REQUIRE `type="permit2"` credentials. The Permit2
 batch transfer mechanism ensures all transfers (primary +
 splits) execute atomically in a single on-chain transaction.
 Servers MUST reject split requests fulfilled with
-`type="transaction"` credentials.
+`type="transaction"`, `type="authorization"`, or
+`type="hash"` credentials.
 
 Each entry in the `splits` array is a JSON object:
 
@@ -552,6 +576,88 @@ This transfers 1.0 USDm to the primary recipient and 0.05 USDm
 to the platform — atomically, in a single transaction. The
 client signs one EIP-712 message covering both transfers.
 
+
+## Authorization Payload (type="authorization") {#authorization-payload}
+
+Opt-in credential type for tokens that implement EIP-3009
+{{EIP-3009}} (e.g., USDC, EURC). The client signs an off-chain
+`transferWithAuthorization` message. The server submits it
+directly to the token contract, paying gas from its own
+balance. The client never interacts with the chain directly.
+
+Unlike `type="permit2"`, this type requires no prior token
+approval — EIP-3009 authorization is built into the token
+contract itself.
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `type` | string | REQUIRED | `"authorization"` |
+| `from` | string | REQUIRED | Payer address |
+| `to` | string | REQUIRED | Recipient address (MUST match challenge `recipient`) |
+| `value` | string | REQUIRED | Transfer amount in base units (MUST match challenge `amount`) |
+| `validAfter` | string | REQUIRED | Unix timestamp — authorization not valid before this time (stringified integer, typically `"0"`) |
+| `validBefore` | string | REQUIRED | Unix timestamp — authorization not valid after this time (stringified integer) |
+| `nonce` | string | REQUIRED | `bytes32` nonce — MUST be set to `challengeHash` (see below) |
+| `signature` | string | REQUIRED | EIP-712 signature (`0x`-prefixed) |
+
+### Challenge Binding
+
+The EIP-3009 `nonce` field is a random `bytes32` value chosen
+by the caller. This specification requires the nonce to be set
+to the `challengeHash`:
+
+~~~
+nonce = keccak256(abi.encodePacked(
+    challenge.id,
+    challenge.realm
+))
+~~~
+
+This provides cryptographic challenge binding equivalent to
+the Permit2 witness mechanism. The token contract enforces
+nonce uniqueness on-chain — a nonce that has been consumed
+cannot be reused, providing replay protection at the contract
+level.
+
+### Constraints
+
+- Splits are NOT supported. Servers MUST reject
+  `type="authorization"` credentials when the challenge
+  includes `splits`.
+- Servers SHOULD only advertise `"authorization"` in
+  `credentialTypes` when the `currency` token is known to
+  implement EIP-3009.
+- `validBefore` SHOULD correspond to the challenge `expires`
+  timestamp.
+- `to` MUST match the challenge `recipient`.
+- `value` MUST match the challenge `amount`.
+
+### Example
+
+~~~json
+{
+  "challenge": {
+    "id": "aB3cDeF4gHiJkLmN",
+    "realm": "api.example.com",
+    "method": "evm",
+    "intent": "charge",
+    "request": "eyJ...",
+    "expires": "2026-04-01T12:05:00Z"
+  },
+  "payload": {
+    "type": "authorization",
+    "from": "0x1234567890abcdef1234567890abcdef12345678",
+    "to": "0x742d35Cc6634C0532925a3b844Bc9e7595f8fE00",
+    "value": "1000000",
+    "validAfter": "0",
+    "validBefore": "1743523500",
+    "nonce": "0x8a3b...f1c2",
+    "signature": "0x..."
+  },
+  "source": "did:pkh:eip155:4326:0x1234567890abcdef1234567890abcdef12345678"
+}
+~~~
+
 ## Transaction Payload (type="transaction") {#transaction-payload}
 
 The compatible fallback. The client signs a complete ERC-20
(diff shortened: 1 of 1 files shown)


Kbhat1:
This is great, thanks for opening the unified one. Much simpler than opening chain specific

0xBreadguy:
> This is great, thanks for opening the unified one. Much simpler than opening chain specific

Thanks! If you see anything you want included I'm all ears for refining. Tried to give rationale for the stepdown approach while incorporating a few features from the SUI/Solana/Tempo specs that I thought were meaningful

gakonst:
Reviewing soon but directionally looks great

kyscott18:
I think this should include an "authorization" credential that uses ERC 3009. It lets the clients skip the step of approving the permit2 contract for tokens that support it (USDC).

Schwartz10:
@0xBreadguy totally agree on this direction! Wondering - is "evm" the right abstraction here? For instance, ignoring implementation complexity for a moment, any given service provider may actually be vm agnostic and care more about currencies than care about receiving a payment over an "evm" machine. For instance, I'm happy to receive USDC on Tempo, Base, or Solana

Not sure the best way to do it, but seems like it might increase usefulness if we can make the abstraction even more general, and work across multiple vms. What do you think?

One approach - I wonder if we could "chain" challenges together, and create an n of m relationship between challenges, kind of like a multisig. I.e. server sends 3 challenges - tempo.charge, base.charge, solana.charge, and requires at least 1 challenge complete. Client can send a credential responding to any challenge, server verifies according to that specific challenge. 

Benefit of this chaining approach is a server can put together more programmatic access to gated content / service by composing individual methods/intents. Server can specify you need to pay AND do action X on smart contract Y in order to receive service. When individual methods can be chained, i think flexibility increases?

0xBreadguy:
> I think this should include an "authorization" credential that uses ERC 3009. It lets the clients skip the step of approving the permit2 contract for tokens that support it (USDC). 

My counter would be that as a universal spec that adds no materially new flow for clients. 

Permit2 captures the single approval flow, is universal to erc20 tokens and can handle the splits feature. 

3009 is an eip-specific implementation of the transfer credential, effectively. 

Not opposed to including it but would want to get consensus from contributors. 

Can also add a note as a candidate for v01 if we do not believe it fits within the MVP.

0xBreadguy:
> @0xBreadguy totally agree on this direction! Wondering - is "evm" the right abstraction here? For instance, ignoring implementation complexity for a moment, any given service provider may actually be vm agnostic and care more about currencies than care about receiving a payment over an "evm" machine. For instance, I'm happy to receive USDC on Tempo, Base, or Solana
> 
> Not sure the best way to do it, but seems like it might increase usefulness if we can make the abstraction even more general, and work across multiple vms. What do you think?
> 

This is what the method layer actually handles, imo. Servers can just issue multiple headers w/ separate methods and the client just selects the one they can fulfill (iiuc):

```
WWW-Authenticate: Payment method="evm", ...
WWW-Authenticate: Payment method="solana", ...
WWW-Authenticate: Payment method="tempo", ...
```
 
Given that, I think `evm` is the right abstraction for methods (and now think `solana` should be converted to `svm` if I'm being honest).

------------
> One approach - I wonder if we could "chain" challenges together, and create an n of m relationship between challenges, kind of like a multisig. I.e. server sends 3 challenges - tempo.charge, base.charge, solana.charge, and requires at least 1 challenge complete. Client can send a credential responding to any challenge, server verifies according to that specific challenge.
> 
> Benefit of this chaining approach is a server can put together more programmatic access to gated content / service by composing individual methods/intents. Server can specify you need to pay AND do action X on smart contract Y in order to receive service. When individual methods can be chained, i think flexibility increases?

As for the challenge bit - super interesting but want to be cognizant of scope creep for what this spec is meant to do, and think that level of change would be more appropriate at the protocol/structural layer as an extension of sorts. I do think it's a useful consideration.

nlok5923:
On gas sponsorship guarantees, Permit2 is recommended partly because "server naturally sponsors gas", but nothing in the spec requires this. What stops a server from advertising `type="permit2"` and then making the client pay gas anyway? Should there be an explicit field in the challenge for this ?

gakonst:
I think would be great to add opt in specialization if 3009 is supported!

gakonst:
And yes EVM is the right abstraction here! The union of methods is of separate concern!

0xBreadguy:
@kyscott18 - Updated to include EIP-3009 as an opt-in, per Georgios rec.

@nlok5923 - updated to be more clear re:gas sponsorship, which is a structural thing for the offchain sigs so doesn't really need explicit fields. lmk if the new note is unclear on reasoning.

ilikesymmetry:
> 3009 is an eip-specific implementation of the transfer credential, effectively.

> Not opposed to including it but would want to get consensus from contributors.

@0xBreadguy I am also in favor of including 3009. Agree with you that Permit2 works for all tokens, but it also requires pre-approval that not all accounts have. 3009 enables a better onboarding experience for tokens that support it and USDC is an important use case today to consider. I don't think including another signature mechanism negatively impacts the devx. It just enables more user choice.

ilikesymmetry:
Otherwise I am in favor of the spec and think it will serve all EVM chains well, great proposal!

gakonst:
Thx folks hopefully we can merge this through the coming week - bit slower weekend due to Easter thx for the understanding 🙌✌️

0xBreadguy:
Updated the initial description to include recently implemented suggestions. 

Will let everyone think on it

specs/methods/evm/draft-evm-charge-00.md
```diff
@@ -0,0 +1,1351 @@
...
+`type="authorization"` (opt-in for EIP-3009 tokens), where
+the client signs an off-chain transfer authorization and
+the server submits it directly to the token contract;
+`type="transaction"`, where the client signs and the server
+broadcasts a standard ERC-20 transfer transaction; and
+`type="hash"` (optional fallback), where the client
+broadcasts the transaction itself and presents the on-chain
+transaction hash for server verification.
+
+This specification covers ERC-20 token transfers only. Native
```
brendanjryan, line 97: 👍

specs/methods/evm/draft-evm-charge-00.md
```diff
@@ -0,0 +1,1351 @@
...
+"charge" intent from {{I-D.payment-intent-charge}}. It defines
+only the EVM-specific `methodDetails`, `payload`, and verification
+procedures.
+
+## Design Rationale
+
+Prior drafts proposed separate payment methods for individual EVM
+chains. However, the control flow, data structures, and
+verification logic are identical across these chains — the only
+differences are chain ID and optional RPC extensions. A unified
```
brendanjryan, line 127: +1

specs/methods/evm/draft-evm-charge-00.md
```diff
@@ -0,0 +1,1351 @@
...
+- **`type="permit2"` (RECOMMENDED)**: The client signs an
+  off-chain EIP-712 Permit2 authorization. The server constructs
+  and submits the on-chain transaction. This is the preferred
+  flow because:
+
+  - The client never interacts with the chain directly
+  - The server naturally sponsors gas (fee payer)
+  - Split payments are atomic via batch transfers
+  - No nonce management burden on the client
+  - `externalId` is cryptographically bound via witness data
```
brendanjryan, line 154: 👍

specs/methods/evm/draft-evm-charge-00.md
```diff
@@ -0,0 +1,1351 @@
...
+
+- **`type="authorization"` (opt-in)**: The client signs an
```
brendanjryan: I don't think we strictly need to mark this as opt-in, only the recommended routes and anything else can be assumed to be optional/opt-in

specs/methods/evm/draft-evm-charge-00.md
```diff
@@ -0,0 +1,1351 @@
...
+20-byte hex strings. Implementations SHOULD use EIP-55
+mixed-case encoding but MUST compare addresses by decoded
+20-byte value, not string form.
+
+## Method Details
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `chainId` | number | REQUIRED | EIP-155 chain ID |
+| `permit2Address` | string | OPTIONAL | Permit2 contract address (default: canonical address) |
```
brendanjryan: should the client need to know the canonical address? IMO it may make more sense just to echo this back every time from the server / on the wire in order to avoid ambiguity

specs/methods/evm/draft-evm-charge-00.md
```diff
@@ -0,0 +1,1351 @@
...
+| `permitted` | array | Array of `{ token, amount }` objects. One entry per transfer (primary + each split). |
+| `nonce` | string | Permit2 nonce (stringified integer) |
+| `deadline` | string | Unix timestamp (stringified integer) |
+
+The `permitted` array MUST always be an array, even for single
+transfers (length 1). Each entry specifies:
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `token` | string | ERC-20 token address (MUST match `currency`) |
```
brendanjryan, line 426: if must match, do we strictly need this?

specs/methods/evm/draft-evm-charge-00.md
```diff
@@ -0,0 +1,1351 @@
...
+  "source": "did:pkh:eip155:4326:0x1234...5678"
+}
+~~~
+
+This transfers 1.0 USDm to the primary recipient and 0.05 USDm
+to the platform — atomically, in a single transaction. The
+client signs one EIP-712 message covering both transfers.
+
+
+## Authorization Payload (type="authorization") {#authorization-payload}
```
brendanjryan, line 592: 👍 I like this modeling as a separate flow which is in _addition_ to permit2, but not required

specs/methods/evm/draft-evm-charge-00.md
```diff
@@ -0,0 +1,1351 @@
...
+
+This provides cryptographic challenge binding equivalent to
+the Permit2 witness mechanism. The token contract enforces
+nonce uniqueness on-chain — a nonce that has been consumed
+cannot be reused, providing replay protection at the contract
+level.
+
+### Constraints
+
+- Splits are NOT supported. Servers MUST reject
```
brendanjryan, line 636: 👍

specs/methods/evm/draft-evm-charge-00.md
```diff
@@ -0,0 +1,1351 @@
...
+  `type="authorization"` credentials when the challenge
+  includes `splits`.
+- Servers SHOULD only advertise `"authorization"` in
```
brendanjryan: why not MUST here? is there a case where we would want to allow this? this seems like a bit of a footgun

specs/methods/evm/draft-evm-charge-00.md
```diff
@@ -0,0 +1,1351 @@
...
+}
+~~~
+
+## Gas Sponsorship Model
+
+Gas sponsorship is structurally determined by the credential
+type — no explicit field is needed:
+
+| Credential Type | Who Pays Gas | Mechanism |
+|----------------|-------------|-----------|
```
brendanjryan, line 768: 👍 this table is helpful

specs/methods/evm/draft-evm-charge-00.md
```diff
@@ -0,0 +1,1351 @@
...
+    "type": "transaction",
+    "signature": "0x02f8...signed transaction bytes..."
+  },
+  "source": "did:pkh:eip155:1329:0x1234567890abcdef1234567890abcdef12345678"
+}
+~~~
+
+# Acknowledgements
+
+The authors thank Georgios Konstantopoulos for guidance on
```
brendanjryan, line 1359: 🙇

specs/methods/evm/draft-evm-charge-00.md
```diff
@@ -0,0 +1,1351 @@
...
+against a different challenge, even if the payment
+parameters are identical.
+
+The witness type string for EIP-712 is:
+`"PaymentWitness witness)PaymentWitness(bytes32 challengeHash)TokenPermissions(address token,uint256 amount)"`
+
+This binding applies to both single and batch
+transfers — the same `witness` parameter is used by
+`permitWitnessTransferFrom()` and
+`permitBatchWitnessTransferFrom()`.
```
brendanjryan, line 484: do you have a reference to this impl? I can't find here: https://docs.uniswap.org/contracts/permit2/reference/signature-transfer

specs/methods/evm/draft-evm-charge-00.md
```diff
@@ -0,0 +1,1351 @@
...
+- The sum of `splits[].amount` MUST be strictly less than
+  `amount`. Clients MUST reject any request that violates
+  this constraint.
+- If present, `splits` MUST contain at least 1 entry.
+  Servers SHOULD limit splits to 10 entries.
+- All transfers MUST target the same `currency` token.
+- Address fields are compared by decoded 20-byte value, not
+  by string form.
+
+The order of entries in `splits` is not significant for
```
brendanjryan: [minor] I think this conflicts with the later Permit2 verification rules. This section says split ordering is not significant, but later verification treats transferDetails[0] as the primary transfer and transferDetails[i+1] as splits[i]. 

Let's choose one one canonical rule here so implementations don’t diverge.

specs/methods/evm/draft-evm-charge-00.md
0xBreadguy: Agreed - should be MUST given authorization==EIP-3009. Will amend.

specs/methods/evm/draft-evm-charge-00.md
0xBreadguy: It can be found here: 
https://docs.uniswap.org/contracts/permit2/reference/signature-transfer#batch-permitwitnesstransferfrom

```
function permitWitnessTransferFrom(
 PermitBatchTransferFrom memory permit,
 SignatureTransferDetails[] calldata transferDetails,
 address owner,
 bytes32 witness,
 string calldata witnessTypeString,
 bytes calldata signature
) external
```

specs/methods/evm/draft-evm-charge-00.md
0xBreadguy: My vote is to make ordering matter throughout to remove ambiguity. 

If you agree, I'll synchronize the spec.

specs/methods/evm/draft-evm-charge-00.md
0xBreadguy: Agreed - put it on the wire in case there are canonical address issues.

Will adjust

specs/methods/evm/draft-evm-charge-00.md
brendanjryan: I'm +1 to this

0xBreadguy: Implemented

specs/methods/evm/draft-evm-charge-00.md
0xBreadguy: Made REQUIRED

specs/methods/evm/draft-evm-charge-00.md
0xBreadguy: No - rationale there was to match the shape of the permit2 call `{ token, amount }`. 

If the redundancy is too much, we can remove to keep the schema lean.

brendanjryan:
Hey! @0xBreadguy -- we think this looks good and could tentatively merge. 

Given the number of prior PRs around this spec + community engagement, would you be open to listing co-authors on the spec?

0xBreadguy:
> Hey! @0xBreadguy -- we think this looks good and could tentatively merge. 
> 
> Given the number of prior PRs around this spec + community engagement, would you be open to listing co-authors on the spec?  

Absolutely.

To remove any personal bias I'd welcome recommendations from the maintainers based on contributions that you saw as meaningful since the original draft.

@brendanjryan

brendanjryan:
@0xBreadguy thank you! I updated the spec with two additional authors who worked on this thread.

Let me know if everything looks good and we will aim to merge by EoW 🙇

0xBreadguy:
Signoff from me @brendanjryan

brendanjryan:
🙇 thanks everybody for your contributions!! very excited to get this merged

brendanjryan (approved):

PR #214, opened 2026-04-01
feat: add NanoPay charge method specification on the TON blockchain

0xterrybit:
NanoPay enables gas-free nanopayments on the TON blockchain via offchain Ed25519 authorization + TEE batch settlement.

- Method name: nanopay
- Intent: charge
- Credential: Ed25519 signed payment authorization
- Verification: TEE aggregator
- Settlement: Batch on-chain (1000:1 compression)
- Minimum payment: $0.000001
- Live service: https://nano.cyberpay.org

### DIFF
--- a/specs/methods/nanopay/draft-nanopay-charge-00.md
+++ b/specs/methods/nanopay/draft-nanopay-charge-00.md
@@ -1,0 +1,236 @@
+---
+title: NanoPay Charge Intent for HTTP Payment Authentication
+abbrev: NanoPay Charge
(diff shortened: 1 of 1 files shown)


brendanjryan:
Hi! Does this offer anything beyond the core EVM mpp spec outlined here? 

https://github.com/tempoxyz/mpp-specs/pull/213

We also would like to avoid adding things like TEEs in formal flows -- these are implementation details and not strictly related to the money movement in this case

PR #201, opened 2026-03-22
Solana session intent specification

lgalabru:
Specification draft for MPP `session` intent, driving an implementation that will the audited and maintained by Solana Foundation (not ready for review)

PR #225, opened 2026-04-06
feat: add unified EVM session payment method

clayclaw:
## Summary

Adds the `session` intent for the `evm` payment method, defining unidirectional streaming payment channels for incremental, voucher-based payments on any EVM-compatible chain. Adapts the streaming payment channel mechanism from the Tempo session spec (`draft-tempo-session`) for general EVM use, with EVM-specific transaction formats, gas models, and domain separators.

- **Request schema**: `amount` (price per unit, base units), `currency` (ERC-20 address), `recipient` (payee address), with `methodDetails` for `chainId`, `escrowContract`, `feePayer`, optional `splits`
- **Escrow contract**: 8 on-chain functions covering channel open, settle, topUp, close, and forced close paths
- **Credential**: four payload actions (`open`, `topUp`, `voucher`, `close`) with EIP-712 voucher signatures for off-chain cumulative payments
- **Verification**: per-action checks — on-chain state for open/topUp, signature recovery + monotonicity for vouchers, final settlement for close
- **Receipt**: per-request `Payment-Receipt` with `acceptedCumulative`, `spent`, optional `reference` (tx hash on settlement)

This is an initial draft, we are actively developing the underlying functionality and will continue refining the spec as implementation progresses. Feedback and discussion are welcome.

### Context

This is the second intent defined for the `evm` payment method (after `charge` in #213). The session intent follows the Tempo session spec (`draft-tempo-session`) as the reference design, extending it to work across any EVM-compatible chain.

### Design decisions

| Decision | Rationale |
|----------|-----------|
| On-chain escrow contract with 8 functions | `open`, `openWithAuthorization`, `settle`, `topUp`, `topUpWithAuthorization`, `close`, `requestClose`, `withdraw` — covers the full channel lifecycle with both client-funded and server-submitted paths. |
| `openWithAuthorization` / `topUpWithAuthorization` via EIP-3009 | When `feePayer: true`, client signs off-chain `transferWithAuthorization`; server calls the escrow contract and pays gas. Client never needs native token. Requires EIP-3009 token support. |
| `open()` + ERC-4337 support for client-broadcast path | When `feePayer: false`, client calls `approve() + open()` directly or via an EntryPoint-mediated UserOperation. Smart wallets can batch both in a single UserOp. |
| Deterministic `channelId` via `keccak256(payer, payee, token, salt, authorizedSigner, escrow, chainId)` | Identical to Tempo's formula. Binds channel to a specific contract deployment and chain. Client can pre-compute before the on-chain tx lands. |
| Deposit merge in voucher payload | Bundles an EIP-3009 `topUp` (or initial `open`) with a voucher update in a single HTTP round-trip — reduces latency when approaching deposit limits. |
| Splits as immutable on-chain `splitRecipients[]` + `splitBps[]` | Session total is unknown upfront, so basis points (not fixed amounts). Enforced atomically by the contract at `settle()` / `close()`. |
| Domain separator: `"EVM Payment Channel"` | Distinguishes from Tempo's `"Tempo Stream Channel"` — the only semantic difference in the voucher signing scheme. Prevents cross-deployment replay. |

## New File

- `specs/methods/evm/draft-evm-session-00.md`

## AI Disclosure

This spec was drafted with AI assistance (Claude). All content has been reviewed for technical accuracy, RFC compliance, and alignment with STYLE.md / CONTRIBUTING.md.

### DIFF
--- a/specs/methods/evm/draft-evm-session-00.md
+++ b/specs/methods/evm/draft-evm-session-00.md
@@ -212,15 +212,17 @@ The following diagrams illustrate the two open modes.
       |  txHash                     |                             |
       |<--------------------------------------------------------------|
       |                             |                             |
-      |  (4) Authorization: Payment |                             |
+      |  (4) Sign initial voucher   |                             |
+      |                             |                             |
+      |  (5) Authorization: Payment |                             |
       |      action="open"          |                             |
       |      type="hash"            |                             |
       |      hash=txHash            |                             |
       |-------------------------->  |                             |
-      |                             |  (5) verify deposit         |
+      |                             |  (6) verify deposit         |
       |                             |-------------------------->  |
       |                             |                             |
-      |  (6) 200 OK + Receipt       |                             |
+      |  (7) 200 OK + Receipt       |                             |
       |<--------------------------  |                             |
       |                             |                             |
       |  ... voucher / close flow (same for both modes) ...       |
--- a/specs/methods/evm/draft-evm-session-00.md
+++ b/specs/methods/evm/draft-evm-session-00.md
@@ -1772,6 +1772,47 @@ possible (e.g., via ERC-4337 UserOperations or
 multicall) to minimize the window between approval and
 channel creation.
 
+## Contract Wallet Signer Mutability {#contract-wallet-signer-mutability}
+
+When `authorizedSigner` is an ERC-1271 contract wallet (e.g., Safe,
+ERC-4337 account), voucher validity depends on the wallet's current
+signer state at verification time, not at signing time. If the wallet's
+owner set, signing key, or signature-validation policy changes after a
+voucher is signed, `isValidSignature` MAY return failure for
+previously-signed vouchers, rendering them unredeemable on-chain.
+
+This creates an asymmetric risk:
+
+- **Payee risk**: If the payer rotates keys on their contract wallet
+  after signing vouchers but before the payee calls `settle()`, the
+  payee loses the ability to redeem accumulated off-chain authorizations.
+- **Payer mitigation**: Payers using contract wallets as
+  `authorizedSigner` SHOULD avoid key rotation during active sessions,
+  or coordinate rotation with settlement.
+- **Payee mitigation**: Payees SHOULD settle more frequently when the
+  `authorizedSigner` is a contract wallet, reducing the value at risk
+  from signer-state changes. Payees MAY inspect the signer address to
+  determine whether it is a contract (via `EXTCODESIZE`) and adjust
+  settlement cadence accordingly.
+
+EOA signers are not affected: ECDSA recovery is stateless and depends
+only on the signature and message.
+
+**Recommended pattern**: When the payer is a contract wallet, the payer
+SHOULD delegate voucher signing to an ephemeral EOA session key by
+setting `authorizedSigner` to that EOA's address, rather than leaving
+`authorizedSigner` unset (which defaults to the payer contract wallet).
+This preserves the AA benefits for the escrowed funds — the contract
+wallet still controls `open()`, `topUp()`, and `close()` calls — while
+eliminating the signer-mutability risk for off-chain vouchers. The
+session key SHOULD be scoped to the lifetime of the channel and
+discarded after `close()`.
+
+Contract-wallet `authorizedSigner` remains permitted for cases where
+EOA delegation is not acceptable (e.g., enterprise multi-sig policies
+that require every signed artifact to carry a quorum signature). In
+such cases, the mitigations above apply.
+
 ## Escrow Guarantees
 
 The escrow contract provides the following security properties:
--- a/specs/methods/evm/draft-evm-session-00.md
+++ b/specs/methods/evm/draft-evm-session-00.md
@@ -186,10 +186,11 @@ authorizations into actual token movement.
 Authorized Signer
 : An address delegated to sign vouchers on behalf of the payer. Defaults
 to the payer if not specified.
-In this specification, voucher signatures are ECDSA
-secp256k1 signatures produced by an EOA-style signer.
-Contract accounts that cannot produce such signatures MUST
-delegate an `authorizedSigner`.
+In this specification, voucher signatures are verified either as
+ECDSA secp256k1 signatures (for EOA signers) or via ERC-1271
+`isValidSignature` (for smart-contract wallets such as Safe or
+ERC-4337 accounts). The `authorizedSigner` field MAY be an EOA
+or an ERC-1271-compliant contract wallet.
 
 Base Units
 : The smallest indivisible unit of an ERC-20 token, determined by the
@@ -327,8 +328,11 @@ All byte arrays (addresses, hashes, signatures, channelId) use:
 
 All signatures in this specification are 65 bytes, encoded as
 `r (32 bytes) || s (32 bytes) || v (1 byte)` and passed as a single
-`bytes` parameter. EIP-2098 compact 64-byte signatures {{EIP-2098}}
-are NOT used; implementations MUST NOT produce or accept them.
+`bytes` parameter. Implementations MUST NOT produce EIP-2098 compact
+64-byte signatures {{EIP-2098}}. Implementations MAY accept them
+(e.g., when using a standard signature-verification library such as
+OpenZeppelin SignatureChecker that transparently handles both
+formats), but MUST NOT require clients to produce them.
 
 Implementations MUST use lowercase hex for channelId, signatures, and
 hashes. Address fields in the request schema (currency, recipient,
@@ -481,14 +485,12 @@ via `receiveWithAuthorization` inside the contract.
 > `receiveWithAuthorization` enforces `msg.sender == to`, preventing
 > front-running attacks where an attacker extracts the EIP-3009
 > signature from the mempool and calls the token directly (see
-> {{front-running-protection}}). The canonical EIP-3009
-> `receiveWithAuthorization` interface accepts separate
-> `(uint8 v, bytes32 r, bytes32 s)` parameters. This escrow contract
-> accepts a packed `bytes` signature instead for interface consistency
-> with `settle` and `close`. The escrow implementation MUST unpack
-> the 65-byte value into `(v, r, s)` components before forwarding to
-> the token's `receiveWithAuthorization`. Tokens that support the
-> USDC v2.2 `bytes signature` overload MAY be called directly.
+> {{front-running-protection}}). This specification targets tokens
+> that support the USDC v2.2 `bytes signature` overload of
+> `receiveWithAuthorization` and calls it directly with the packed
+> 65-byte signature. Tokens that only expose the canonical
+> `(uint8 v, bytes32 r, bytes32 s)` interface are NOT supported by
+> this escrow design.
 
 | Parameter | Type | Description |
 |-----------|------|-------------|
@@ -576,6 +578,7 @@ behalf of the payer.
 |-----------|------|-------------|
 | `channelId` | bytes32 | Existing channel identifier |
 | `additionalDeposit` | uint128 | Additional amount |
+| `topUpSalt` | bytes32 | Random value for nonce derivation |
 | `from` | address | Payer address |
 | `validAfter` | uint256 | EIP-3009 validity start |
 | `validBefore` | uint256 | EIP-3009 validity end |
@@ -586,6 +589,7 @@ behalf of the payer.
 function topUpWithAuthorization(
     bytes32 channelId,
     uint128 additionalDeposit,
+    bytes32 topUpSalt,
     address from,
     uint256 validAfter,
     uint256 validBefore,
@@ -603,11 +607,16 @@ the settlement delta is distributed according to the split ratios
 (same logic as `settle`), then the remaining deposit is refunded to
 the payer.
 
+If `cumulativeAmount <= channel.settled`, the payee is forfeiting
+any uncollected amount (e.g., to cleanly close an exhausted or
+abandoned channel). In this case the contract MAY skip voucher
+signature verification and `signature` MAY be empty.
+
 | Parameter | Type | Description |
 |-----------|------|-------------|
 | `channelId` | bytes32 | Channel to close |
 | `cumulativeAmount` | uint128 | Final cumulative amount |
-| `signature` | bytes | EIP-712 voucher signature |
+| `signature` | bytes | EIP-712 voucher signature; MAY be empty when `cumulativeAmount <= channel.settled` (forfeit path) |
 
 ~~~solidity
 function close(
@@ -669,8 +678,8 @@ functions that accept voucher signatures (`settle`, `close`):
    signer address from the EIP-712 signature and verify it matches:
     - `channel.authorizedSigner` if non-zero
     - Otherwise `channel.payer`
-      Contract accounts that cannot produce secp256k1 ECDSA signatures
-      MUST configure `authorizedSigner` to an EOA-style signer.
+      The signer may be an EOA (verified via ECDSA) or an ERC-1271
+      contract wallet (verified via `isValidSignature`).
 
 3. **Domain binding**: The contract MUST use its own address as
    `verifyingContract` in the EIP-712 domain separator, ensuring
@@ -1053,6 +1062,7 @@ REQUIRED, as described in the Credential Structure section.
 | `action` | string | REQUIRED | `"topUp"` |
 | `type` | string | REQUIRED | `"transaction"` |
 | `channelId` | string | REQUIRED | Channel ID |
+| `salt` | string | REQUIRED | Random bytes32 hex for nonce derivation |
 | `authorization` | object | REQUIRED | EIP-3009 authorization parameters |
 | `signature` | string | REQUIRED | EIP-3009 signature |
 | `additionalDeposit` | string | REQUIRED | Additional amount to deposit |
@@ -1082,7 +1092,7 @@ atomically.
 | `deposit.action` | string | REQUIRED | `"open"` or `"topUp"` |
 | `deposit.authorization` | object | REQUIRED | EIP-3009 authorization parameters (type, from, to, value, validAfter, validBefore, nonce) |
 | `deposit.signature` | string | REQUIRED | EIP-3009 signature (65 bytes, hex-encoded) |
-| `deposit.salt` | string | CONDITIONAL | Random bytes32 hex for channelId computation. REQUIRED when `deposit.action` is `"open"`; MUST NOT be present for `"topUp"` |
+| `deposit.salt` | string | REQUIRED | Random bytes32 hex. Used for channelId computation when `deposit.action` is `"open"`; passed as `topUpSalt` for nonce derivation when `deposit.action` is `"topUp"` |
 | `deposit.authorizedSigner` | string | OPTIONAL | Address delegated to sign vouchers. Defaults to payer (`authorization.from`) if omitted. Only applicable when `deposit.action` is `"open"` |
 
 When `deposit` is present, the server processes the deposit first
@@ -1141,6 +1151,7 @@ if `open` is called on an existing `channelId`.
     "signature": "0xabcdef...vouchersig",
     "deposit": {
       "action": "topUp",
+      "salt": "0xcccc5678dddd9012eeee3456ffff7890aaaa1234bbbb5678cccc9012dddd3456",
       "authorization": {
         "type": "eip-3009",
         "from": "0xaabbccddee11223344556677889900aabbccddee",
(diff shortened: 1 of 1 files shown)


specs/methods/evm/draft-evm-session-00.md
```diff
@@ -0,0 +1,1997 @@
...
+version: 00
+category: info
+ipr: trust200902
+submissiontype: IETF
+consensus: true
+
+author:
+  - name: Xin Tian
+    ins: X. Tian
+    email: xin.tian@okg.com
```
brendanjryan, line 14: does this email resolve? domain 404s

specs/methods/evm/draft-evm-session-00.md
```diff
@@ -0,0 +1,1997 @@
...
+
+--- middle
+
+# Introduction
+
+This document is published as Informational but contains normative
+requirements using BCP 14 keywords {{RFC2119}} {{RFC8174}} to ensure
+interoperability between implementations.
+
+The `session` intent is an **experimental intent** defined in
```
brendanjryan: I'm ok with keeping this for now, but by defining a new generic EVM intent we may want to consider formalizing things. We (Tempo Labs) will drive this

specs/methods/evm/draft-evm-session-00.md
```diff
@@ -0,0 +1,1997 @@
...
+|-------|------|-------------|
+| `payer` | address | User who deposited funds |
+| `payee` | address | Server authorized to withdraw |
+| `token` | address | {{ERC-20}} token address |
+| `authorizedSigner` | address | Authorized signer (0 = payer) |
+| `deposit` | uint128 | Total amount deposited |
+| `settled` | uint128 | Cumulative amount already withdrawn by payee |
+| `closeRequestedAt` | uint64 | Timestamp when close was requested (0 if not) |
+| `finalized` | bool | Whether channel is closed |
+| `splitRecipients` | address[] | Split recipient addresses (empty if no splits) |
```
brendanjryan: is there a note on why we want to include splits here? i'm not strictly opposed to this, but it does make the settlement process more complicated as you need to split on each settle

specs/methods/evm/draft-evm-session-00.md
```diff
@@ -0,0 +1,1997 @@
...
+| `splitBps` | uint16[] | Corresponding basis points per recipient |
```
brendanjryan: have you considered gas costs at all for this structure? we have optimized structs and storage on tempo, and this will be worse on other EVM chains. Not required but something to consider

specs/methods/evm/draft-evm-session-00.md
```diff
@@ -0,0 +1,1997 @@
...
+) external returns (bytes32 channelId);
+~~~
+
+When `splitRecipients` is empty, the channel has no splits and
+all settlement funds go to the payee. Split parameters are
+immutable once the channel is created.
+
+### openWithAuthorization
+
+Opens a channel using EIP-3009 {{EIP-3009}} authorization. The server
```
brendanjryan, line 531: have you considered other methods in addition to EIP-3009? I think this works well for some stablecoins but not globally

specs/methods/evm/draft-evm-session-00.md
```diff
@@ -0,0 +1,1997 @@
...
+
+1. **Voucher acceptance**: Verify and persist new `acceptedCumulative`
+2. **Balance check**: If `available < cost`, return 402
+3. **Charge and deliver**: Persist `spent := spent + cost` BEFORE
+   delivering service
+4. **Receipt generation**: Include balance state
+
+## Crash Safety
+
+- Persist `spent` increments BEFORE delivering service
```
brendanjryan, line 1978: 👍

specs/methods/evm/draft-evm-session-00.md
```diff
@@ -0,0 +1,1997 @@
...
+voucher updates and content streaming.
+
+For voucher-only updates (no response body needed), clients MAY use
+`HEAD` requests.
+
+## Receipt Generation {#receipt-generation}
+
+Servers MUST return a `Payment-Receipt` header on **every successful
+paid request**. For streaming responses (SSE), servers MUST include the
+receipt in the initial response headers AND as a final SSE event:
```
brendanjryan, line 2106: 👍

brendanjryan:
Directionally looks good -- thanks for putting this together!

I think one thing to consider is support for non-EIP3009 tokens via permit or similar. 

This isn't _strictly_ required at this time, but we should make sure the design will scale to this model without breaking the core interface

specs/methods/evm/draft-evm-session-00.md
clayclaw: Yes the email address is working

specs/methods/evm/draft-evm-session-00.md
clayclaw: Yes, we also plan to consider supporting Permit2

specs/methods/evm/draft-evm-session-00.md
clayclaw: Agreed. At the protocol level I think it make sense to align with existing implementations. That would also make it easier to define a more standard session intent.

brendanjryan:
Two issues that should be fixed before this lands:

1. Stale voucher handling currently returns `200 OK` for `cumulativeAmount <= highestVoucherAmount` before verifying the voucher signature/signer. 

Since voucher credentials can omit `source` and identify the payer from channel state, anyone who knows a `channelId` could submit a stale or low unsigned voucher and potentially consume already-authorized balance. Please verify signer first, or only treat an exact stored credential/signature replay as idempotent.

2. The EIP-3009 channel-parameter binding needs to be mandatory. 

The draft correctly notes that EIP-3009 does not bind `payee`, `salt`, or `authorizedSigner`, but deterministic nonce derivation is only a SHOULD. Because `openWithAuthorization` is callable by anyone, a relayer that sees the authorization could front-run with attacker-controlled channel parameters. Please make the deterministic nonce derivation a MUST and require the escrow to reject a supplied nonce that does not match the derived value, or remove the nonce argument and derive it internally.

brendanjryan:
cc @clayclaw -- let me know when this is ready fro another review!

chopmob-cloud:
+1 on the unified EVM approach. We ship the x402 v2 canonical envelope identically across Base and Tempo (USDC on both) with zero per-chain branching at the HTTP layer, so the design intent of "one session spec, multiple EVM chains" matches what works in production.

A few operational notes from running both chains live (AlgoVoi gateway):

- **Revert short-circuit**: `eth_getTransactionReceipt status=0x0` should be handled as a typed error rather than waiting for the spinner timeout. Worth calling out as a verifier obligation so session-`close` paths don't hang on a reverted settlement.
- **Deterministic `channelId` binding**: including `chainId` in the keccak formula prevents the cross-deployment replay you'd otherwise hit if a payer reused the same authorized-signer across two EVMs. We use the same pattern on Recurr Tier 2 EVM authorities and it simplifies replay reasoning significantly.
- **Fee-absorption tolerance on voucher monotonicity**: gas-price spikes on Base can cause `topUp` to land for less than the client expected after slippage. Our CCTP V2 xchain flow uses a `tolerated_min_delivered` field to absorb the 14 bps Fast Transfer fee. An analogous tolerance on the voucher `cumulativeAmount` boundary might be worth specifying so channels don't force-close on a thin margin.
- **On @brendanjryan's EIP-3009 vs permit concern**: this matches our experience that USDC's EIP-3009 coverage is the happy path; for non-3009 tokens our Recurr Tier 2 EVM module falls back to `approve()` + transfer with the user-broadcast path, which is the same shape as the `feePayer: false` branch here. The interface should scale to permit-based and approve-based tokens without breaking the core abstraction.

Happy to dig into any of these if useful for the spec text.

chopmob-cloud:
Coming at this from the non-EVM side of the same problem space — AlgoVoi ships AP2 (Agent Payment Protocol) mandates as the Algorand/VOI equivalent of the session intent defined here, so some of the design decisions look familiar. Happy to contribute production observations as this develops.

Two points that may be useful as the spec is refined:

**On EIP-3009 universality.** brendanjryan's point is well-placed. On Algorand/VOI we hit the same gap: not all assets support the off-chain-authorisation pattern, so we run a fallback path where the client pre-funds an escrow and the server draws down against it without needing per-call authorisation. Permit2 is the right EVM analogue — worth specifying the fallback path explicitly in the spec rather than leaving it to implementers, otherwise you get fragmentation between EIP-3009-capable and non-capable token deployments.

**On splits at settlement.** `splitBps[]` enforced at `settle()` / `close()` is the right shape — we run the same pattern in production. One edge case worth covering: what happens if a split recipient is a contract that reverts on receive? A push-based split at settlement can deadlock the channel close. A pull-based alternative (split recipients claim their share post-settlement) avoids the revert-griefing vector entirely.

Both are refinement points rather than blockers. The domain separator distinction from Tempo's `"Tempo Stream Channel"` is correct practice. We're following this thread and glad to help stress-test the spec further as implementation progresses — cross-chain production experience with the same intent pattern is a useful reference point.

*AlgoVoi (chopmob-cloud) — Acquisition enquiries: https://docs.algovoi.co.uk/acquisition*

clayclaw:
@brendanjryan Thanks, both issues are addressed in the latest draft. Voucher idempotency now runs only after EIP-712 signature and signer verification, and the EIP-3009 channel-parameter binding is now mandatory.

@chopmob-cloud Thanks for the production notes, we added explicit `receipt.status == 0x0` handling with a typed transaction-reverted error so settle/close don’t hang on reverts. On voucher cumulative-amount tolerance for gas/top-up slippage, we’re leaning toward keeping that chain-specific rather than normative here, and we’re open to community discussion if others want a normative hook.

In this revision we also split the escrow into a mandatory core and an optional relayed / gasless operations profile for relayer-submitted paths (EIP-3009, Permit2, payee-authorized settle/close), with advertised capabilities aligned to what the escrow actually implements.

Settlement amounts must strictly increase: the escrow reverts stale vouchers on-chain, and servers apply the same rule off-chain. Finalized channels stay on record permanently, with a cross-epoch replay section, so reusing the same channelId cannot revive old vouchers against a new deposit.

clayclaw:
> Coming at this from the non-EVM side of the same problem space — AlgoVoi ships AP2 (Agent Payment Protocol) mandates as the Algorand/VOI equivalent of the session intent defined here, so some of the design decisions look familiar. Happy to contribute production observations as this develops.
> 
> Two points that may be useful as the spec is refined:
> 
> **On EIP-3009 universality.** brendanjryan's point is well-placed. On Algorand/VOI we hit the same gap: not all assets support the off-chain-authorisation pattern, so we run a fallback path where the client pre-funds an escrow and the server draws down against it without needing per-call authorisation. Permit2 is the right EVM analogue — worth specifying the fallback path explicitly in the spec rather than leaving it to implementers, otherwise you get fragmentation between EIP-3009-capable and non-capable token deployments.
> 
> **On splits at settlement.** `splitBps[]` enforced at `settle()` / `close()` is the right shape — we run the same pattern in production. One edge case worth covering: what happens if a split recipient is a contract that reverts on receive? A push-based split at settlement can deadlock the channel close. A pull-based alternative (split recipients claim their share post-settlement) avoids the revert-griefing vector entirely.
> 
> Both are refinement points rather than blockers. The domain separator distinction from Tempo's `"Tempo Stream Channel"` is correct practice. We're following this thread and glad to help stress-test the spec further as implementation progresses — cross-chain production experience with the same intent pattern is a useful reference point.
> 
> _AlgoVoi (chopmob-cloud) — Acquisition enquiries: https://docs.algovoi.co.uk/acquisition_

Thanks for the AP2 context. We just added non–EIP-3009 funding path with Permit2 in the relayed profile, with feePayerAuthorizations for discovery. We had split / splitBps in an earlier draft but removed them from the unified escrow surface: revenue splits feel better as an implementor extension than core session semantics, especially with extra gas at settle/close that varies by chain. This also adds the settlement complexity. Your push-vs-pull revert point is still a useful note for anyone layering splits on top. Thanks for sharing these notes.

chopmob-cloud:
The channel mechanics are well-structured — the deterministic `channelId`, EIP-712 voucher monotonicity, and the `openWithAuthorization` / `topUpWithAuthorization` split across fee-payer paths cover the core EVM constraints cleanly.

One gap worth addressing before the spec stabilises: the settlement receipt on `close` / `settle` is currently thin — `acceptedCumulative`, `spent`, and an optional `reference` (tx hash). This doesn't distinguish terminal states as positive records. A verifier looking at a channel history cannot tell the difference between a clean `close` (mutual agreement), a `requestClose` timeout path, and a forced `withdraw` — they all produce the same receipt shape, or no receipt at all if the tx hash isn't present.

The pattern that closes this: a terminal state enum (`SETTLED / PENDING_FINALITY / REVERSED`) carried in the settlement receipt alongside a content-addressed `settled_payment_ref` and a `canon_version` pin. The verifier recomputes the receipt hash from canonical bytes and can distinguish final settlement from forced close from reversal as byte-distinct records — without querying the escrow contract state at audit time.

This is specified in [`draft-hopley-x402-settlement-attestation`](https://datatracker.ietf.org/doc/draft-hopley-x402-settlement-attestation/) and runs in production across AlgoVoi's EVM payment flows today. The reference implementation is at [`@algovoi/settlement-attestation`](https://www.npmjs.com/package/@algovoi/settlement-attestation) (Apache 2.0). Happy to cross-reference the spec if it's useful for the settlement receipt shape here.

*AlgoVoi (chopmob-cloud) — [docs.algovoi.co.uk/settlement-attestation](https://docs.algovoi.co.uk/settlement-attestation)*

specs/methods/evm/draft-evm-session-00.md
```diff
@@ -0,0 +1,2614 @@
...
+`openWith…()` function on the escrow contract. The
+`authorization.type` field selects the format.
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `action` | string | REQUIRED | `"open"` |
+| `type` | string | REQUIRED | `"transaction"` |
+| `channelId` | string | REQUIRED | Channel identifier (hex bytes32) |
+| `authorization` | object | REQUIRED | Token-pull authorization parameters; format determined by `authorization.type` |
+| `signature` | string | REQUIRED | Authorization signature (65 bytes hex). EIP-3009 signature if `authorization.type="eip-3009"`; Permit2 EIP-712 signature if `authorization.type="permit2"` |
```
brendanjryan: **Credential `type` vocabulary diverges from the sibling `evm` method (`draft-evm-charge`)**

Both specs are `method="evm"`, so a client's evm credential decoder is shared — but the `type` discriminator means different things across the two intents:

- `type="transaction"` here = "server submits a token-pull authorization." In `draft-evm-charge` it means "client signs a *complete* ERC-20 transfer tx; server broadcasts." Same string, inverted responsibility.
- EIP-3009 is `authorization.type="eip-3009"` here, but `type="authorization"` (top-level) in charge.
- Permit2 is `authorization.type="permit2"` (nested) here, but `type="permit2"` (top-level) in charge.

**Suggested fix:** reuse charge's top-level `type` names and drop the nested `authorization.type`. Session uses a subset (no "full signed tx" path, so it omits `transaction`):

| `type` (shared across `evm`) | Who submits | Client signs |
|---|---|---|
| `permit2` | server | Permit2 `SignatureTransfer` |
| `authorization` | server | EIP-3009 authorization |
| `hash` | client | broadcasts; submits tx hash |

So the feePayer:true open payload becomes `type="authorization"` or `type="permit2"` (not `type="transaction"` + a nested `authorization.type`); `type="hash"` already matches charge. Keep the `authorization` sub-object for the format-specific params, but select its shape from the top-level `type`. Apply the same rename to the TopUp payload and the `voucher` deposit-merge (`deposit.type` instead of `deposit.authorization.type`).

specs/methods/evm/draft-evm-session-00.md
```diff
@@ -0,0 +1,2614 @@
...
+## Method Details
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `methodDetails.chainId` | number | REQUIRED | EVM chain ID |
+| `methodDetails.escrowContract` | string | REQUIRED | Address of the channel escrow contract |
+| `methodDetails.channelId` | string | OPTIONAL | Channel ID if resuming an existing channel |
+| `methodDetails.minVoucherDelta` | string | OPTIONAL | Minimum amount increase between vouchers (base units). Default: `"0"` (any positive increment accepted). See {{dos-mitigation}} |
+| `methodDetails.feePayer` | boolean | OPTIONAL | If `true`, server pays gas for open/topUp (default: `false`) |
+| `methodDetails.feePayerAuthorizations` | array | CONDITIONAL | Authorization formats the server and its escrow support, as an ordered list of `authorization.type` values (`"eip-3009"`, `"permit2"`). REQUIRED when `feePayer` is `true`; omitted/ignored otherwise. Order expresses server preference |
```
brendanjryan: **`feePayerAuthorizations` duplicates `credentialTypes` from `draft-evm-charge`**

`draft-evm-charge` already negotiates supported credential formats via `methodDetails.credentialTypes` (ordered preference list). This introduces a second, overlapping field for the same job in the same `evm` method. Once the format names are aligned (see the credential-`type` comment), this field is redundant.

**Suggested fix:** drop `feePayerAuthorizations` and reuse `credentialTypes`:

```
methodDetails.credentialTypes: ["permit2", "authorization", "hash"]   // same semantics as draft-evm-charge
```

brendanjryan (commented):
Inline review of two cross-spec consistency issues with the other evm/session specs on paymentauth.org (see comments below).

brendanjryan:
@clayclaw -- left a few more comments, but this looks pretty close! 

Along with https://github.com/tempoxyz/mpp-specs/pull/201, we are going to try to formalize the generic "session" intent this week, which should unblock this pr from merging.

brendanjryan:
https://github.com/tempoxyz/mpp-specs/pull/280

Here is the generic session spec, can you look at the above and flag any deltas or misalignment?

clayclaw:
> #280
> 
> Here is the generic session spec, can you look at the above and flag any deltas or misalignment?

Thanks. Let me take a look

clayclaw:
> #280
> 
> Here is the generic session spec, can you look at the above and flag any deltas or misalignment?

We reviewed the generic session intent PR and aligned the EVM session draft where it makes sense.

Apart from the IANA cleanup, the main alignment change was to receipt semantics.

Change:
- Aligned receipt semantics: `reference` is now the stable session reference and equals `channelId`; on-chain transaction hashes are returned separately as optional `txHash` on settlement/close receipts.

One note: we did not change EVM’s `insufficient-balance` problem type to the generic draft’s `insufficient-value`. Tempo already uses `https://paymentauth.org/problems/session/insufficient-balance`, EVM matches that, and the currently working URI appears to be the balance one. Can you confirm if this is a possible inconsistency in the generic draft rather than something method specs should all change?

clayclaw:
> The channel mechanics are well-structured — the deterministic `channelId`, EIP-712 voucher monotonicity, and the `openWithAuthorization` / `topUpWithAuthorization` split across fee-payer paths cover the core EVM constraints cleanly.
> 
> One gap worth addressing before the spec stabilises: the settlement receipt on `close` / `settle` is currently thin — `acceptedCumulative`, `spent`, and an optional `reference` (tx hash). This doesn't distinguish terminal states as positive records. A verifier looking at a channel history cannot tell the difference between a clean `close` (mutual agreement), a `requestClose` timeout path, and a forced `withdraw` — they all produce the same receipt shape, or no receipt at all if the tx hash isn't present.
> 
> The pattern that closes this: a terminal state enum (`SETTLED / PENDING_FINALITY / REVERSED`) carried in the settlement receipt alongside a content-addressed `settled_payment_ref` and a `canon_version` pin. The verifier recomputes the receipt hash from canonical bytes and can distinguish final settlement from forced close from reversal as byte-distinct records — without querying the escrow contract state at audit time.
> 
> This is specified in [`draft-hopley-x402-settlement-attestation`](https://datatracker.ietf.org/doc/draft-hopley-x402-settlement-attestation/) and runs in production across AlgoVoi's EVM payment flows today. The reference implementation is at [`@algovoi/settlement-attestation`](https://www.npmjs.com/package/@algovoi/settlement-attestation) (Apache 2.0). Happy to cross-reference the spec if it's useful for the settlement receipt shape here.
> 
> _AlgoVoi (chopmob-cloud) — [docs.algovoi.co.uk/settlement-attestation](https://docs.algovoi.co.uk/settlement-attestation)_

Thanks for the pointer. The session drafts seem to use a lightweight receipt model today: accounting state plus method-native evidence like txHash.

The proposed attestation format looks broader than this EVM session draft, and adding it only here may make EVM diverge from the other session methods. This may fit better as a shared receipt extension across session methods later.

brendanjryan (approved):
Thank you for the revisions! This LGTM

PR #226, opened 2026-04-08
[intent] authorize intent (auth and capture)

brendanjryan:
## Summary

Adds the `authorize` intent as a full authorization-and-capture lifecycle for HTTP Payment Authentication, with method-specific capture, void, and refund-request semantics.

## Key design decisions

* Supports escrowed auth and capture flows, common with cards and many other payment methods
* Allows for multi-capture as well as single-capture
* Supports "voiding" a transaction, but *not* partial refunds

## Tempo Implementation

Leverages the same underlying "escrow" contract as the tempo session method, with the server as an authorized signer

## Stripe Implementation

Leverages shared payment token, with restrictions on what parameters and shapes of payment methods can be used.

## Open questions

* should we allow partial refunds? 
* are there any other efficiencies that we could get from the tempo contract?

### DIFF
--- a/specs/intents/draft-payment-intent-authorize-00.md
+++ b/specs/intents/draft-payment-intent-authorize-00.md
@@ -47,35 +47,40 @@ normative:
 
 This document defines the "authorize" payment intent for use with the
 Payment HTTP Authentication Scheme. The "authorize" intent represents a
-pre-authorization where the payer grants the server permission to charge
-up to a specified amount within a time window, without immediate payment.
+payment authorization where the payer approves a maximum amount that a
+server can later capture before an expiry time.
 
 --- middle
 
 # Introduction
 
-The "authorize" intent enables pre-authorized payments where the payer
-grants the server permission to charge up to a specified amount at a
-later time. This is useful for:
+The "authorize" intent enables delayed payment capture. The payer
+authorizes a maximum amount, and the payment method creates a hold,
+escrow, or equivalent authorization. The server can later capture one or
+more amounts against that authorization, subject to the authorized maximum
+and expiry.
 
-Metered billing:
-: Pay-per-use APIs where total cost is unknown upfront
+This is useful for:
 
 Delayed fulfillment:
-: Services where delivery occurs after authorization
+: Services where delivery or shipment occurs after payment authorization.
+
+Metered billing:
+: Services where final cost is unknown when the payer authorizes payment.
 
 Spending caps:
-: User-controlled limits on automated spending
+: User-controlled limits on future server-initiated captures.
 
-Unlike the "charge" intent which requires immediate payment, "authorize"
-creates a payment capability that the server can exercise later.
+Unlike the "charge" intent, successful authorization is not itself a
+payment capture. A successful authorization response MUST NOT imply that
+the recipient has received funds.
 
 ## Relationship to Payment Methods
 
-Payment methods implement "authorize" using method-specific
-authorization mechanisms. This document defines the abstract semantics
-and shared request fields; payment method specifications define how
-those semantics are enforced.
+This document defines the abstract authorize semantics and shared request
+fields. Payment method specifications define how an authorization is
+created, how captures are executed, how unused authorizations are voided,
+and which method-specific policy applies to refund requests.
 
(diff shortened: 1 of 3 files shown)


evchip:
1.  > Funds availability: With authorization keys on tempo, this method just gives users the "right" to capture funds, but does not guarantee funds can be captured (requires simulation). This makes this not a true auth+capture flow. To address this we could use a similar (same?) contract to what we use for sessions

Simulation is not a complete guarantee of funds availability. Escrow contract is needed to enable the 'without immediate payment' feature, else it must be capture-first-deliver-after, but that forces per-charge onchain settlement. It also negates the delayed fulfillment use case (server should not start provisioning unless payment is guaranteed).

2. Authorize is functionally different from charge and session in that there is nothing preventing servers from overcharging. Charge and session intents both require client signatures for each transaction/voucher, but authorize allows servers to charge anything up to the cap, regardless of actual resource consumption. IMO best to explicitly note the loss of per-charge consent so implementers understand what guarantees authorize is/isn't giving them.

brendanjryan:
@evchip 

> Escrow contract is needed to enable the 'without immediate payment' feature, else it must be capture-first-deliver-after, but that forces per-charge onchain settlement.

Correct -- this is one tradeoff with this design and something we do not like either. We are going to do another rev of this which uses an escrow, which also allows for partial + multi-capture flows -- an ask we have seen from users in the last few weeks. 

We expect to have a new update to this in ~2 weeks!

specs/intents/draft-payment-intent-authorize-00.md
```diff
@@ -0,0 +1,669 @@
...
+ipr: noModificationTrust200902
+submissiontype: IETF
+consensus: true
+
+author:
+  - name: Jake Moxey
+    ins: J. Moxey
+    email: jake@tempo.xyz
+    org: Tempo Labs
+  - name: Brendan Ryan
```
brendanjryan, line 16: authors TBD

specs/intents/draft-payment-intent-authorize-00.md
```diff
@@ -0,0 +1,669 @@
...
+      |                              |----------------------------->|
+      |                              |                              |
+      |  (3) 200 OK                  |  (capture confirmed)          |
+      |      Payment-Receipt         |<-----------------------------|
+      |<-----------------------------|                              |
+      |                              |                              |
+~~~
+
+The core intent deliberately does not expose payment method capture
+capabilities, such as whether the method performs one capture or multiple
```
brendanjryan, line 179: important note!

specs/intents/draft-payment-intent-authorize-00.md
```diff
@@ -0,0 +1,669 @@
...
+          |<---------------------------|
+          |                            |
+~~~
+
+Void closes the authorization and releases any uncaptured value according
+to method-specific rules. Void does not refund captured value.
+
+## Refund Requests
+
+Refunds are out of scope for the core authorize lifecycle. Clients MAY
```
brendanjryan, line 201: note

specs/methods/tempo/draft-tempo-authorize-00.md
```diff
@@ -0,0 +1,790 @@
...
+ipr: noModificationTrust200902
+submissiontype: IETF
+consensus: true
+
+author:
+  - name: Jake Moxey
+    ins: J. Moxey
+    email: jake@tempo.xyz
+    org: Tempo Labs
+  - name: Brendan Ryan
```
brendanjryan, line 16: authors TBD

specs/methods/stripe/draft-stripe-authorize-00.md
```diff
@@ -0,0 +1,672 @@
...
+abbrev: Stripe Authorize
+docname: draft-stripe-authorize-00
+version: 00
+category: info
+ipr: noModificationTrust200902
+submissiontype: IETF
+consensus: true
+
+author:
+  - name: Brendan Ryan
```
brendanjryan, line 12: authors tbd!

brendanjryan:
closing this out as it is stale

PR #230, opened 2026-04-08
Subscriptions: Intent + Stripe and Tempo Implementations

brendanjryan:
## Summary

This PR adds the `subscription` payment intent for recurring fixed-amount Payment authentication, plus Stripe and Tempo method profiles.

The shared intent defines activation, renewal, cancellation, subscription identifiers, canonical period semantics, durable accounting/idempotency, receipts, and error behavior.

Stripe maps the intent to a constrained Stripe Billing profile. Tempo maps it to TIP-1011 periodic access-key authorizations.

## Notable design decisions / tightening of spec

* The shared intent is intentionally narrow: fixed amount, one charge per billing period. It does not model full billing-product behavior like plans, seats, prorations, trials, discounts, metered usage, or plan changes.

* Payment methods must reject request shapes they cannot represent exactly. Shared periods support fixed elapsed day/week periods and calendar-month periods anchored at activation. Stripe supports exact day/week/month cadence; Tempo supports day/week and rejects month because TIP-1011 periods are fixed elapsed seconds.

* Activation includes the first billing-period charge. A subscription is not active until setup and the first charge succeed, and the receipt includes a `subscriptionId`.

* Renewals require durable per-period accounting. Missed periods do not accumulate extra charge authority, and duplicate requests, retries, webhooks, or concurrent requests must not produce duplicate charges.

* Stripe is constrained to one customer, one subscription, one recurring price, quantity 1, synchronous first-invoice payment, and validated paid invoices mapped to canonical billing periods.

* Tempo requires bounded authorization with `subscriptionExpires`, exact `TokenLimit` amount/period matching, recipient-scoped transfer selectors, source verification, and access-key isolation guidance.

## Control flow

```mermaid
flowchart TD
  A["Client requests protected resource"] --> B{"Active subscription for resource?"}
  B -- "yes, current period paid" --> C["Return receipt + resource"]
  B -- "yes, new unpaid period" --> D["Collect renewal via method profile"]
  D --> E["Durably record paid period"]
  E --> C
  B -- "no / expired / canceled / revoked" --> F["Return 402 subscription challenge"]
  F --> G["Client signs method-specific subscription credential"]
  G --> H["Server verifies challenge, request, payer, method scope"]
  H --> I{"Payment method"}
  I -- "Stripe" --> J["Create/reuse constrained Stripe subscription + paid first invoice"]
  I -- "Tempo" --> K["Register scoped access key + transfer first charge"]
  J --> L["Record subscription state + billing anchor"]
  K --> L
  L --> M["Return Payment-Receipt with subscriptionId"]

### DIFF
--- a/specs/intents/draft-payment-intent-subscription-00.md
+++ b/specs/intents/draft-payment-intent-subscription-00.md
@@ -110,8 +110,8 @@ Subscription
   billing period.
 
 Billing Period
-: A fixed-duration window during which at most one subscription charge
-  may be collected.
+: A recurrence window, defined by the subscription period fields, during
+  which at most one subscription charge may be collected.
 
 Activation
 : The successful initial setup of a subscription, which includes
@@ -203,24 +203,36 @@ document those choices explicitly.
 |-------|------|-------------|
 | `amount` | string | Fixed payment amount per billing period in base units |
 | `currency` | string | Currency or asset identifier (see {{currency-formats}}) |
-| `periodSeconds` | string | Billing period duration in seconds |
+| `periodUnit` | string | Billing period unit. The value MUST be `day`, `week`, or `month` |
+| `periodCount` | string | Positive integer count of `periodUnit` values per billing period |
 
 The `amount` value MUST be a string representation of a positive
 integer in base 10 with no sign, decimal point, exponent, or
 surrounding whitespace. Leading zeros MUST NOT be used.
 
-The `periodSeconds` value MUST be a string representation of a positive
+The `periodCount` value MUST be a string representation of a positive
 integer in base 10 with no sign, decimal point, exponent, or
 surrounding whitespace. Leading zeros MUST NOT be used.
 
-`periodSeconds` defines fixed-duration billing periods measured in
-elapsed seconds. It does not, by itself, encode calendar-month or
-calendar-year alignment.
+The `periodUnit` value defines how billing-period boundaries are
+computed:
 
-Payment methods MUST reject request objects whose `periodSeconds` value
-they cannot represent exactly. They MUST NOT approximate the requested
-period by rounding, truncating, or substituting a nearby network-native
-cadence.
+- `day`: fixed elapsed-time periods of `periodCount * 86400` seconds
+- `week`: fixed elapsed-time periods of `periodCount * 604800` seconds
+- `month`: calendar-month periods anchored at activation
+
+For `periodUnit="month"`, billing-period boundaries are computed by
+adding `N * periodCount` calendar months to the original activation
+anchor using UTC calendar fields. If the target month does not contain
+the activation anchor's day-of-month, the boundary uses the last valid
+day of the target month while preserving the activation time-of-day.
+Boundaries MUST be computed from the original activation anchor, not by
+adding months to the previous boundary.
+
+Payment methods MUST reject request objects whose `periodUnit` or
+`periodCount` values they cannot represent exactly. They MUST NOT
+approximate the requested period by rounding, truncating, or
+substituting a nearby network-native cadence.
 
 ### Optional Fields
 
@@ -255,8 +267,8 @@ so.
 
 The billing anchor for a subscription is the time activation succeeds,
 or an equivalent network-native timestamp defined by the payment method
-specification. Billing periods are contiguous fixed-duration windows
-derived by adding `periodSeconds` to that anchor.
+specification. Billing periods are contiguous windows derived from that
+anchor according to `periodUnit` and `periodCount`.
 
 This shared intent does not define deferred starts or merchant-selected
 billing anchors. A payment method that needs a more specific anchor rule
@@ -264,9 +276,10 @@ MUST document it explicitly.
 
 The shared fields in this section are the canonical subscription
 contract. Payment method specifications MUST document how they map
-`amount`, `periodSeconds`, and activation to the underlying payment
-system. If a payment method cannot represent those fields or semantics
-exactly, it MUST reject the request rather than approximate it.
+`amount`, `periodUnit`, `periodCount`, and activation to the underlying
+payment system. If a payment method cannot represent those fields or
+semantics exactly, it MUST reject the request rather than approximate
+it.
 
 ## Currency Formats {#currency-formats}
 
@@ -302,8 +315,8 @@ In particular:
 
 - Methods should support only request shapes they can represent exactly.
 - Methods should document the supported and rejected ranges or values of
-  `periodSeconds`, any additional bounded-lifetime or expiry rules they
-  impose, and what conditions make activation succeed.
+  `periodUnit` and `periodCount`, any additional bounded-lifetime or
+  expiry rules they impose, and what conditions make activation succeed.
 - Activation should not be reported as successful until both
   subscription setup and the first billing-period charge have
   succeeded.
@@ -327,7 +340,8 @@ In particular:
 {
   "amount": "9900",
   "currency": "usd",
-  "periodSeconds": "2592000",
+  "periodUnit": "month",
+  "periodCount": "1",
   "description": "Pro plan"
 }
 ~~~
@@ -338,7 +352,8 @@ In particular:
 {
   "amount": "10000000",
   "currency": "0x20c0000000000000000000000000000000000001",
-  "periodSeconds": "2592000",
+  "periodUnit": "day",
+  "periodCount": "30",
   "subscriptionExpires": "2026-07-14T12:00:00Z",
   "recipient": "0x742d35cc6634c0532925a3b844bc9e7595f8fe00",
   "methodDetails": {
@@ -488,26 +503,34 @@ usable and initiate a new subscription flow.
 
 This section is non-normative.
 
-## 30-Day Billing Example
+## Monthly Billing Example
 
 Suppose a server offers a plan with these request fields:
 
 - `amount = "9900"`
 - `currency = "usd"`
-- `periodSeconds = "2592000"`
+- `periodUnit = "month"`
+- `periodCount = "1"`
 
 If activation succeeds at `2026-01-15T12:03:10Z`, that time becomes the
 billing anchor. The resulting billing periods are:
 
-- Period 0: `[2026-01-15T12:03:10Z, 2026-02-14T12:03:10Z)`
-- Period 1: `[2026-02-14T12:03:10Z, 2026-03-16T12:03:10Z)`
-- Period 2: `[2026-03-16T12:03:10Z, 2026-04-15T12:03:10Z)`
+- Period 0: `[2026-01-15T12:03:10Z, 2026-02-15T12:03:10Z)`
+- Period 1: `[2026-02-15T12:03:10Z, 2026-03-15T12:03:10Z)`
+- Period 2: `[2026-03-15T12:03:10Z, 2026-04-15T12:03:10Z)`
 
 Activation collects the Period 0 charge. Requests during Period 0 do
 not require another renewal charge. When Period 1 begins, the server
 may collect one renewal charge for Period 1 before, or atomically with,
 granting access for that period. After that renewal succeeds, additional
 requests during Period 1 do not permit another charge for Period 1.
+
+If the same monthly plan activates at `2026-01-31T12:03:10Z`, the
+resulting boundaries are computed from the January 31 anchor:
+
+- Period 0: `[2026-01-31T12:03:10Z, 2026-02-28T12:03:10Z)`
+- Period 1: `[2026-02-28T12:03:10Z, 2026-03-31T12:03:10Z)`
+- Period 2: `[2026-03-31T12:03:10Z, 2026-04-30T12:03:10Z)`
 
 ## Cancellation Example
 
@@ -559,7 +582,7 @@ Clients MUST verify before activating a subscription:
 
 1. `amount` is acceptable for the service
 2. `currency` is expected
-3. `periodSeconds` matches the expected billing interval
+3. `periodUnit` and `periodCount` match the expected billing interval
 4. Any `subscriptionExpires` value and method-specific constraints are
    understood and acceptable
 
--- a/specs/intents/draft-payment-intent-subscription-00.md
+++ b/specs/intents/draft-payment-intent-subscription-00.md
@@ -418,8 +418,13 @@ subscription namespace.
 Clients MAY retain the `subscriptionId` as application data when
 referring to the active subscription in later interactions.
 
-This specification does not define a dedicated HTTP request header for
-carrying `subscriptionId`.
+The `Subscription-Id` header is only a subscription-selection hint. It
+does not, by itself, prove authority to use the subscription.
+
+The challenge `request` object does not carry `subscriptionId`. The
+canonical reusable identifier is the `subscriptionId` field returned in
+the `Payment-Receipt`, and the corresponding request-time selector is
+the `Subscription-Id` header.
 
 Servers MUST authenticate or otherwise authorize the client's use of the
 identified subscription before granting access or collecting a renewal
--- a/specs/intents/draft-payment-intent-subscription-00.md
+++ b/specs/intents/draft-payment-intent-subscription-00.md
@@ -114,8 +114,8 @@ Billing Period
   may be collected.
 
 Activation
-: The successful initial registration of a subscription, which also
-  collects the first billing-period charge.
+: The successful initial setup of a subscription, which includes
+  collection of the first billing-period charge.
 
 Renewal
 : A later charge that collects the subscription amount for a subsequent
@@ -126,8 +126,8 @@ Cancellation
 
 Subscription Identifier
 : A server-issued opaque identifier for an activated subscription,
-  used by clients to re-authenticate into that subscription on later
-  requests.
+  used by servers and applications to refer to that subscription in
+  later interactions.
 
 # Intent Semantics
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: specs/methods/tempo/draft-tempo-subscription-00.md [hunk 1]
```diff
@@ -0,0 +1,511 @@
...
+    "signature": "0xf8c1...signed authorization bytes...",
+    "type": "keyAuthorization"
+  },
+  "source": "did:pkh:eip155:42431:0x1234567890abcdef1234567890abcdef12345678"
+}
+~~~
+
+## Unsupported Payload Types
+
+Tempo subscriptions do not support `type="transaction"` or
```
brendanjryan: why do we need to define this? there are not any types in these txns

specs/methods/tempo/draft-tempo-subscription-00.md
```diff
@@ -0,0 +1,511 @@
...
+1. Verify `amount` is reasonable for the service
+2. Verify `currency` is the expected TIP-20 token address
+3. Verify `periodSeconds` matches expectations
+4. Verify `recipient` is controlled by the expected party
+5. Verify `subscriptionExpires` is acceptable
+
+## Revocation
+
+Users can revoke subscription access keys at any time via the
+AccountKeychain precompile. Servers SHOULD handle revocation gracefully
```
brendanjryan: include a reference here

specs/methods/tempo/draft-tempo-subscription-00.md
```diff
@@ -0,0 +1,511 @@
...
+by returning a fresh subscription challenge.
+
+## Duplicate Charge Prevention
+
+On-chain periodic limits prevent overspending within a billing period,
```
brendanjryan, line 535: include a reference here

specs/intents/draft-payment-intent-subscription-00.md
```diff
@@ -0,0 +1,583 @@
...
+    date: 2026-01
+  I-D.ietf-httpapi-idempotency-key-header:
+    title: "The Idempotency-Key HTTP Header Field"
+    target: https://datatracker.ietf.org/doc/draft-ietf-httpapi-idempotency-key-header/
+    author:
+      - name: Jayadeba Jena
+    date: 2024-06
+---
+
+--- abstract
```
tarasmitran-stripe, line 46: I'm curious if there is a doc on the motivation?  Eg, why build subscriptions into the tempo layer vs utilize a billing system and then create payment on Tempo?  Not a statement of judgement, just trying to understand the product goals.

specs/intents/draft-payment-intent-subscription-00.md
```diff
@@ -0,0 +1,583 @@
...
+supported subset.
+
+### Required Fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `amount` | string | Fixed payment amount per billing period in base units |
+| `currency` | string | Currency or asset identifier (see {{currency-formats}}) |
+| `periodSeconds` | string | Billing period duration in seconds |
+| `subscriptionExpires` | string | Subscription expiry timestamp in {{RFC3339}} format |
```
tarasmitran-stripe: why does a subscription need to expire?  Eg, if you sign up for a service like Netflix there is not expiration

specs/intents/draft-payment-intent-subscription-00.md
```diff
@@ -0,0 +1,583 @@
...
+REQUIRED in their method specification, and MUST document any narrower
+supported subset.
+
+### Required Fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `amount` | string | Fixed payment amount per billing period in base units |
+| `currency` | string | Currency or asset identifier (see {{currency-formats}}) |
+| `periodSeconds` | string | Billing period duration in seconds |
```
tarasmitran-stripe: what is the start date?  Often subscriptions do not start now() but anchored to something like UTC midnight.

specs/intents/draft-payment-intent-subscription-00.md
```diff
@@ -0,0 +1,583 @@
...
+preserve the semantics in this document exactly or reject the request;
+it MUST NOT approximate them.
+
+# Requirements Language
+
+{::boilerplate bcp14-tagged}
+
+# Terminology
+
+Subscription
```
tarasmitran-stripe, line 108: This is a fairly narrow definition, that does not match the billing perspective, which might be fine, but who is the audience for this concept and would they be confused?  

One other thought - given the big definition difference between this subscription and a billing subscription - should we consider another name (maybe 'recurring')?

specs/intents/draft-payment-intent-subscription-00.md
```diff
@@ -0,0 +1,583 @@
...
+MUST be strictly later than the challenge `expires` timestamp. Servers
+MUST reject credentials where `subscriptionExpires` is at or before the
+challenge `expires`.
+
+The first billing period begins when the subscription is activated.
+Payment methods MAY define additional activation controls in
+`methodDetails`, but MUST define exact activation semantics if they do
+so.
+
+The billing anchor for a subscription is the time activation succeeds.
```
tarasmitran-stripe: services even such as Netflix support resetting the billing cycle anchor - is that a trapdoor decision here?

specs/intents/draft-payment-intent-subscription-00.md
```diff
@@ -0,0 +1,583 @@
...
+All payment methods implementing the "subscription" intent MUST support
+these shared fields. Payment methods MAY elevate OPTIONAL fields to
+REQUIRED in their method specification, and MUST document any narrower
+supported subset.
+
+### Required Fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `amount` | string | Fixed payment amount per billing period in base units |
```
tarasmitran-stripe, line 204: most subscription systems do not take a fixed amount, but rather a price and quantity, and compute the amount.  Eg, you are buying 5 seats to a SaaS system

specs/intents/draft-payment-intent-subscription-00.md
brendanjryan: that's fair -- I think we could have this be optional and only have a firm expiry if desired 

alternative we could remove and add later as an optional field to reduce API surface

specs/intents/draft-payment-intent-subscription-00.md
brendanjryan: Recurring is interesting -- that may be more semantically correct as this is a charge x interval. Will think about this a bit more

specs/intents/draft-payment-intent-subscription-00.md
brendanjryan: this is defined later -- but start date _should_ be immediately, this could have some variability at the payment method (e..g stripe, tempo) level though

specs/methods/stripe/draft-stripe-subscription-00.md
```diff
@@ -0,0 +1,618 @@
...
+
+# Settlement Procedure
+
+## Activation and First-Period Charge
+
+For `intent="subscription"`, the server MUST:
+
+1. Create or reuse a Stripe Customer for the payer
+2. Attach or select the challenged `paymentMethod` for that Customer
+3. Create or reuse a Stripe Price whose amount, currency, and recurring
```
tarasmitran-stripe, line 370: What about the Stripe Product?

specs/methods/stripe/draft-stripe-subscription-00.md
```diff
@@ -0,0 +1,618 @@
...
+Payers MUST be able to cancel Stripe subscriptions. For this profile,
+the default cancellation effective time is the end of the current paid
+canonical billing period.
+
+When a payer cancels, the server MUST set the Stripe Subscription to
+cancel at the period end corresponding to the last paid canonical
+billing period, and MUST record that cancellation effective time in
+durable local state. The server MAY cancel immediately only if the
+application separately handles any already-paid access period without
+creating an additional charge.
```
tarasmitran-stripe, line 463: Can you un-cancel before period end?

specs/methods/stripe/draft-stripe-subscription-00.md
```diff
@@ -0,0 +1,618 @@
...
+- no subscription schedule
+
+Servers MUST create the subscription using an idempotency key bound to
+the challenge ID, payer, payment method, amount, currency, and
+`periodUnit` and `periodCount`. If an idempotent retry returns an
+existing Stripe Subscription, the server MUST verify that the existing
+object still matches this profile before treating the retry as
+successful.
+
+## Unsupported Stripe Billing Features
```
tarasmitran-stripe, line 291: What did we decide would happen if the user goes and modifies the stripe subscription and adds one of these features?

specs/methods/stripe/draft-stripe-subscription-00.md
brendanjryan: to keep things simple for now let's say no

specs/methods/stripe/draft-stripe-subscription-00.md
brendanjryan: at this point we would consider this "out of protocol" and something that I think could be disputed along the given rail (this is easier to do in stripe vs. in crypto which is strictly defined by the protocol)

specs/methods/stripe/draft-stripe-subscription-00.md
brendanjryan: IIRC Stripe requires a Price to belong to a Product, but in this case, product selection does not affect the protocol authorization. 

I think we can treat product as an implementation detail here? or we could also specify to be very explicit if desired

chopmob-cloud:
Operational notes from shipping multi-chain MPP subscriptions to production (AlgoVoi gateway, deployed 2026-05-13). Adding the data because the spec is well-shaped and we hit the edge cases the spec strictness is designed to prevent.

## What we shipped

Subscription intent live across 7 chains (Algorand, VOI, Hedera, Stellar, Base, Solana, Tempo) under migrations 109-112. Implementation surface: `mpp_subscriptions` table, `subscriptions.py` router, `subscription_payment_hook.py` hook, dashboard subscription view. Resource declarations carry `is_subscription` + period fields. The intent shape we landed maps cleanly onto this PR's shared intent — same fixed-amount-per-period, same activation-includes-first-charge, same canonical period semantics.

## Where the spec strictness paid off in production

**1. "Activation includes the first billing-period charge."** Strong yes from operational data. We tried the alternative (separate "setup" and "first charge") in early iterations and hit a ~3% subscription-state-mismatch rate where the merchant got an activation event but the on-chain transfer reorged or failed asynchronously, leaving the customer with an "active" subscription that never actually paid. Folding activation into the first-charge atomicity eliminates that window entirely.

**2. "Renewals require durable per-period accounting."** This is the hardest requirement to implement correctly across chains because the finality assumptions differ. We landed a `payment_ledger_chain` table (migration 110) that anchors each renewal to a verified on-chain settlement before the period is marked paid. Specifically:

- Algorand / VOI: 4-round confirmation (~16s) before marking paid
- Solana: 1-slot finality treated as paid; full finalization is monitored async
- Hedera: HCS consensus timestamp + HTS allowance match
- Stellar: Soroban auth_entry expiry tracking
- Base / Tempo: 12-block depth (~3min) for "paid" status

The spec text "missed periods do not accumulate extra charge authority" maps cleanly to a `period_authorisation_high_water_mark` semantic we enforce per subscription. If a chain has a reorg past a paid period, the period flips back to unpaid and the renewal flow re-fires; if a payer comes back after a 3-period gap, only one renewal can be charged (the current period), not three.

**3. "Duplicate requests, retries, webhooks, or concurrent requests must not produce duplicate charges."** The strongest operational case for the spec text. We use a JCS-canonical hash of `(subscriptionId, billing_period_index)` as a dedup key — same canonicalization rule we're locking into the parallel x402 spec work in [x402-foundation/x402#2334](https://github.com/x402-foundation/x402/pull/2334) under `JCS_hash`. Cross-protocol consistency: an MPP subscription renewal and an x402-anchored receipt for that same renewal hash identically, so a downstream aggregator deduplicates within and across protocols on one rule.

## Edge cases worth surfacing in the spec text

These aren't suggested changes — the current strictness already prevents the failure modes — but they're worth surfacing in non-normative implementation notes:

**a) "Synchronous first-invoice payment" requirement vs chain finality.** The spec mandates the first-invoice charge succeeds before activation. For chains with subjective finality (Solana 1-slot, EVM 1-block), implementations need to decide what "succeeded synchronously" means at the protocol layer. We treat 1-slot/1-block as "soft success" for the activation flow + an async monitor that flips the subscription to revoked if the source tx is dropped before configured finality depth. Worth noting in the spec text that the per-method profile defines what depth qualifies as "success" rather than baking a single rule.

**b) Calendar-month vs fixed-elapsed period semantics on long-cycle subscriptions.** The spec correctly limits Stripe to "exact day/week/month cadence" and Tempo to "day/week (rejects month)" because TIP-1011 periods are fixed elapsed seconds. We hit this when porting subscriptions to Algorand: ASA-based subscriptions use Algorand's `LastValid` round-deadline pattern which is fixed elapsed rounds (≈ fixed elapsed time at ~3.3s/round). A calendar-month subscription can't be expressed as fixed rounds without per-month re-issuance. We landed a `period_type` column with two values (`elapsed_fixed` and `calendar_anchored`) — merchants pick at resource registration, and the method profile rejects the combination it can't represent. Same pattern the spec describes for Tempo rejecting month.

**c) `subscriptionId` uniqueness across chains and methods.** Worth tightening (or noting non-normatively) that `subscriptionId` should be method-namespaced or globally unique across the merchant's resource registry. We had a brief incident where a Stripe-method subscription and a Tempo-method subscription on the same resource generated collidable IDs at the receipt-store level. Fixed by prefixing with `method:` (`stripe:cs_test_abc` / `tempo:tip1011_xyz`). Probably already handled by Stripe's `cus_*` / `sub_*` namespacing and TIP-1011's `subscriptionExpires` key, but worth flagging for new method profiles like the one in #258 (Hedera).

**d) Chain-portable activation receipts.** Our `Payment-Receipt` envelope carries `subscriptionId` plus `chain_id` plus the on-chain tx hash, all under one JCS-canonical signed envelope. This means a single receipt can be cross-verified against the on-chain rail regardless of which chain processed it — useful for the cross-class composition row in #2322 / #2334. Not normative for this PR but worth noting that the receipt shape the spec defines is rail-portable as long as the method profile maps cleanly to a deterministic on-chain anchor.

## Cross-PR coordination

The subscription intent here, the x402 PaymentRequired body shape in [x402-foundation/x402](https://github.com/x402-foundation/x402), and the AP2 PaymentMandate in [google-agentic-commerce/AP2](https://github.com/google-agentic-commerce/AP2) all hash receipts under SHA-256(JCS(...)) when we ship them. Three protocols, one canonicalization rule. Worth noting in the spec text (non-normatively) that an MPP subscription receipt and an x402 receipt for the same underlying settlement should hash identically under JCS so cross-protocol aggregators deduplicate cleanly. This is the rule the working group locked into across [#2322](https://github.com/x402-foundation/x402/pull/2322) and [#2334](https://github.com/x402-foundation/x402/pull/2334) on the x402 side.

## Composability with the parallel Hedera session intent in #258

We support Hedera as a payment chain and have a Hedera-side subscription implementation that maps onto HTS Approve allowances. Happy to bring operational data to that PR if useful — the activation-includes-first-charge requirement here interacts non-trivially with HTS allowance issuance order (allowance must be on-chain before the first charge can hit). Worth a coordination note between #230 and #258 on what the activation atomicity looks like when the underlying primitive is an allowance rather than a direct transfer.

PR reads ready for review from our side; happy to engage further on any of these points if useful for the spec maintainer.

specs/methods/stripe/draft-stripe-subscription-00.md
tarasmitran-stripe: Yes, its an implementation detail in general, but it does create a user-facing entity since the price requires the product to exist

chopmob-cloud:
Following up on my 2026-05-16 comment with a correction.

Bullet 1 ("activation includes the first billing-period charge") described our intended design, not what was deployed at the time. The other operational notes were accurate as shipped, including the per-period accounting + missed-period non-accumulation invariants and the JCS dedup approach.

As of today the gateway implements the full atomic-activation flow per spec §Activation L396-407: authority tx verified inline, first-period pull executed on chain (broadcast-ACK soft success), subscription created with `first_pull_tx_id` captured, async finality monitor flips the subscription to a new `revoked` state on chain drop or finality timeout. `200` with `status: "active"` and a `Payment-Receipt` header is returned only after the first-pull is broadcast-accepted.

Today's commits on `chopmob-cloud/AlgoVoi-Hand` master:

| Spec area | Commit |
|---|---|
| subscriptionId base64url + Payment-Receipt header (§Subscription Identifier) | `8522e0aa` |
| Calendar-anchored month + missed-period skip-past (§Request Schema, §Renewal) | `cb48be0b` |
| Single-use credential / reject replay (§Single-Use) | `30dbad6d` |
| Idempotency-Key per draft-ietf-httpapi-idempotency-key-header §2.5 (L469-472) | `ec601997` |
| Synchronous first-period charge + revoked state (§Activation L396-407) | `c296f3a2` |

All five live on `api.algovoi.co.uk` now. Full per-credential activation against a funded wallet is the operational next step; happy to coordinate a cross-vendor test against the live `mpp-sub-test-daily-base` fixture once funding is in place.

brendanjryan:
Merging this in as this has now been implemented in Tempo, with Stripe support to follow. 

We will open incremental PRs if there are any changes needed

parvahuja (approved):

PR #231, opened 2026-04-09
spec: add Accept-Payment negotiation

brendanjryan:
## Summary

Many MPP servers support multiple methods/intents, which is good for increasing coverage of compatible clients, but can result in wasted computation on the server (generating details for methods that a client may never use) as well as unclear selection criteria in clients, which today use the _first supported_ method returned by the server

We address this by adding an `Accept-Payment` header, which allows clients to declare which payment methods they support and their relative preference.

This is modeled after the `Accept-Language` header 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Language

### DIFF
--- a/specs/core/draft-httpauth-payment-00.md
+++ b/specs/core/draft-httpauth-payment-00.md
@@ -565,9 +565,9 @@ intent-or-wildcard         = intent-token / "*"
 Examples:
 
 ~~~http
-Accept-Payment: tempo/charge, tempo/session, stripe/charge;q=0.5
-Accept-Payment: tempo/*, */session;q=0.3
-Accept-Payment: tempo/charge, tempo/session;q=0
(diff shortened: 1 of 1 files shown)


specs/core/draft-httpauth-payment-00.md
```diff
@@ -545,6 +545,54 @@ WWW-Authenticate: Payment id="def", realm="api.example.com", method="example", i
...
+payment-range  = payment-token [ weight ]
+payment-token  = payment-method-or-wildcard "/" intent-or-wildcard
+payment-method-or-wildcard = payment-method-id / "*"
+intent-or-wildcard         = intent-token / "*"
+~~~
+
+Examples:
+
+~~~http
+Accept-Payment: tempo/charge, tempo/session, stripe/charge;q=0.5
```
lgalabru: ❤️

jxom (approved):

tmm (approved):

PR #247, opened 2026-04-16
[draft] Add EVM stake intent

Schwartz10:
### Context

My team and I have been building tools for untrusted agentic coordination. A primitive we’ve found useful is refundable stake: a client locks collateral to access a resource, gets it back if it behaves as expected, and can be slashed otherwise. This is a simple way to align behavior and deter abuse without requiring an irreversible payment.

A couple of our internal projects use this pattern today, so I wrote this draft for MPP in case it is useful to others. In some deployments, the locked collateral can also be routed into yield while it sits, which opens up new types of business models.

### For the community:

1. [Stake-to-pay app kit](https://github.com/gitbondhq/stake-to-pay) 
 
Developers can easily fork this template and build "stake to pay" apps over MPP. This includes a basic smart contract, server implementation, a CLI, and agent skills. 

2. [mppx-stake](https://www.npmjs.com/package/@gitbondhq/mppx-stake) - an npm package that implements the "stake" intent spec 

3. Live demo - GitBond Gatekeeper - a "stake-to-open-pull-request" on GitHub, currently in alpha:  [GitHub App](https://gh.gitbond.com/setup), [smart contracts](https://github.com/gitbondhq/escrow).

Requires a small refundable stake before an external contributor can open a PR on GitHub. Try it out by opening a PR here: https://github.com/gitbondhq/pr-me

The goal is to illustrate how "staking" is a useful MPP primitive by making low-effort or spammy AI-generated PRs more costly, while keeping the bond refundable for legitimate contributors. The UX is still rough; this is meant to show the mechanism in practice.

AI disclosure: AI helped turn our working code prototypes into this draft’s spec prose and format, but it did not originate the underlying mechanism design. 

Feedback welcome!

### DIFF
--- a/specs/methods/evm/draft-evm-stake-00.md
+++ b/specs/methods/evm/draft-evm-stake-00.md
@@ -1,0 +1,797 @@
+---
+title: EVM Stake Intent for HTTP Payment Authentication
+abbrev: EVM Stake
+docname: draft-evm-stake-00
+version: 00
+category: info
+ipr: noModificationTrust200902
+submissiontype: IETF
+consensus: true
+
+author:
+  - name: Jonathan Schwartz
+    ins: J. Schwartz
+    email: jonathan@glif.io
+    org: GitBond
+
+normative:
+  RFC2119:
+  RFC3339:
+  RFC4648:
+  RFC8174:
+  RFC8259:
+  EIP-712:
+    title: "Typed structured data hashing and signing"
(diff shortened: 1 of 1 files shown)


brendanjryan:
Thanks for submitting!

I think this is a very interesting pattern, but not something we would want to enshrine in the core protocol at this time, as it does not generalize well outside of crypto. 

My suggestion would be to have this as its own payment method and distribute it to services that want to use it

PR #255, opened 2026-05-04
spec: add currency and network parameters to Accept-Payment

### LINKED ISSUE #254: Accept-Payment: add currency and network preference parameters
## Problem

The `Accept-Payment` header (Section 7.4) currently negotiates only `method/intent` combinations. A client can express "I prefer `evm/charge` over `stripe/charge`" but **cannot** express:

- "I hold USDC on Base, not Ethereum L1"
- "I accept USDC but not EURC"
- "I prefer chain 8453 over chain 56"

All currency and network information is locked inside the base64url-encoded `request` parameter. A client receiving multiple challenges must decode every `request` blob to discover what currency and network each challenge requires — then discard the ones it can't fulfill.

As EVM chain variants proliferate (Base, BNB, Arbitrum, and others) and multi-rail servers become common, this is a **correctness and efficiency requirement**, not a convenience feature. An agent with a USDC wallet on Base shouldn't need to decode and discard challenges for Ethereum L1, Solana devnet, and three card networks before finding the one it can pay.

## Proposed Solution

Extend `payment-range` with optional semicolon-separated preference parameters, following the same pattern as HTTP `Accept` media type parameters (RFC 9110).

### Proposed ABNF

```abnf
Accept-Payment = #payment-range
payment-range  = payment-token *prefer-param [ weight ]
payment-token  = payment-method-or-wildcard
                 "/" intent-or-wildcard
prefer-param   = OWS ";" OWS token "=" ( token / quoted-string )
```

### Two Initial Parameters

**`currency`** — the currency or asset the client can pay with. Format is method-scoped: ISO 4217 lowercase for fiat methods (`usd`), token contract address for on-chain methods (`0x833589fC...`), or method-defined identifiers (`sat` for Lightning).

**`network`** — the payment network or chain. Format is method-scoped: EIP-155 chain ID for EVM (`8453`), CAIP-2 for Stellar (`stellar:pubnet`), network name for Solana/Lightning (`mainnet`), or omitted for methods without a network concept (Stripe, card).

### Example

```http
Accept-Payment: evm/charge;network=8453;currency=0x833589fC,
  solana/charge;network=mainnet;currency=EPjFWdd5Au;q=0.8,
  lightning/charge;network=mainnet;q=0.5,
  stripe/charge;currency=usd;q=0.3
```

The client prefers USDC on Base, then USDC on Solana, then Lightning, then Stripe.

## Design Decisions

| Decision | Rationale |
|----------|-----------|
| Generic `*prefer-param` syntax, not dedicated ABNF fields | Future parameters (`credential-type`, `max-amount`, `fee-payer`) can be added without touching core ABNF |
| Parameters are method-scoped (opaque strings in core spec) | Each payment method already uses different currency/network formats — forcing a single format would break existing conventions |
| Absent parameter = match any | Full backward compatibility. Existing `Accept-Payment` headers work unchanged |
| **Unknown parameters MUST be ignored** | Forward compatibility. New parameters can be added in future drafts without breaking current implementations — established HTTP evolution pattern (Accept, Cache-Control, Content-Type) |
| Case-sensitive byte-for-byte matching | Consistent with how `method` and `intent` are already matched |
| Ranges with parameters are more specific | Follows HTTP content negotiation precedent |

## Spec Sections Affected

- **Section 7.4** — ABNF, new "Preference Parameters" subsection, updated server matching logic
- **Appendix A** — ABNF collected (must match Section 7.4)
- **Appendix B** — new examples (currency/network preferences, cross-method, partial)
- **Security Considerations** — new "Accept-Payment Information Disclosure" subsection
```


### Additional context

## Cross-Protocol Context

No other agent payment protocol has solved client-side payment preference signaling:

| Protocol | Client pre-signals? | Limitation |
|----------|-------------------|------------|
| **MPP (this spec)** | Yes — `Accept-Payment` | Only method/intent today |
| **x402** | No | Server dictates via `accepts[]` |
| **ACP** | Partial | `capabilities` object, but no payment method preference |
| **UCP** | Partial | Profile URL, but broad capabilities only |
| **FADP** | No | Single challenge, take-it-or-leave-it |

Extending `Accept-Payment` with currency/network makes it the first cross-method, cross-chain payment preference mechanism for HTTP.

## Implementation

A PR with the complete spec change is being opened shortly from [arditti/mpp-specs@accept-payment-currency-network](https://github.com/arditti/mpp-specs/tree/accept-payment-currency-network). Happy to iterate on the design in the PR review.

### PR DESCRIPTION
**Pull request by arditti**
Closes #254

## Summary

Extends Section 7.4 (Client Payment Preferences) of the core spec to support optional `currency` and `network` parameters in the `Accept-Payment` header, enabling clients to express fine-grained payment preferences beyond `method` and `intent`.

This mirrors the issue discussion at #254 — currency and network are currently locked inside the base64url-encoded `request` parameter, forcing clients to decode every challenge to find one they can fulfill.

### ABNF extension (Section 7.4 + Appendix A)

`payment-range` gains a `*prefer-param` production that allows semicolon-separated `key=value` pairs between the `method/intent` token and the `q` weight:

```abnf
payment-range  = payment-token *prefer-param [ weight ]
prefer-param   = OWS ";" OWS token "=" ( token / quoted-string )
```

### Two initial parameters

- **`currency`** — ISO 4217 lowercase for fiat, token contract address for on-chain, or method-defined identifier (`sat` for Lightning)
- **`network`** — EIP-155 chainId for EVM, CAIP-2 for Stellar, network name for Solana/Lightning, omitted for methods without a network concept

Both are method-scoped: valid values are defined by each method's specification. The core spec treats them as opaque strings.

### Server matching logic

Extended the "servers SHOULD" list in Section 7.4 to intersect `currency` and `network` when filtering challenges, with a specificity rule (ranges with parameters are more specific than ranges without).

### Examples

Added three new examples in Appendix B:
- Currency and Network Preferences (single-method filter)
- Cross-Method Preferences (EVM + Solana + Lightning + Stripe in one header)
- Partial Preferences (currency without network, or vice versa)

### Security consideration

New "Accept-Payment Information Disclosure" subsection notes that clients who want to avoid disclosing wallet capabilities can omit the header and filter locally.

## Design Decisions

| Decision | Rationale |
|----------|-----------|
| Generic `*prefer-param` syntax | Future parameters (`credential-type`, `max-amount`, `fee-payer`) added without core ABNF changes |
| Method-scoped opaque strings | Each method already uses different formats — no single canonical format exists |
| Absent parameter = match any | Backward compatibility — existing `Accept-Payment` headers work unchanged |
| Unknown parameters MUST be ignored | Forward compatibility — established HTTP evolution pattern |
| Case-sensitive byte-for-byte matching | Consistent with how `method` and `intent` are matched |

## Backward Compatibility

Fully backward compatible. Existing `Accept-Payment` headers without parameters continue to work exactly as before. A range without a parameter matches any value for that dimension.

## Validation

- `make check` — passed (kramdown-rfc + xml2rfc + rfclint)
- `make lint` — passed (frontmatter)
- `python3 scripts/lint_external_section_refs.py` — passed (no new `Section X.Y of {{I-D...}}` refs)
- All modified lines under 72 chars

## AI Assistance Disclosure

Per CONTRIBUTING.md: I used Claude to help draft the spec prose and validate the ABNF. All content has been reviewed for RFC compliance, technical accuracy against the existing spec structure, and consistency with the STYLE.md design principles.

## Files Changed

- `specs/core/draft-httpauth-payment-00.md` — single-file change across Section 7.4, Appendix A, Appendix B, and Security Considerations

### DIFF
--- a/specs/core/draft-httpauth-payment-00.md
+++ b/specs/core/draft-httpauth-payment-00.md
@@ -556,8 +556,10 @@ negotiation fields: omitted `q` values are equivalent to `q=1`, and
 
 ~~~abnf
 Accept-Payment = #payment-range
-payment-range  = payment-token [ weight ]
-payment-token  = payment-method-or-wildcard "/" intent-or-wildcard
+payment-range  = payment-token *prefer-param [ weight ]
+payment-token  = payment-method-or-wildcard
+                 "/" intent-or-wildcard
+prefer-param   = OWS ";" OWS token "=" ( token / quoted-string )
 payment-method-or-wildcard = payment-method-id / "*"
 intent-or-wildcard         = intent-token / "*"
 ~~~
@@ -915,10 +917,12 @@ payment-credentials = "Payment" 1*SP base64url-nopad
 
 ; Client payment preferences
 Accept-Payment = #payment-range
-payment-range = payment-token [ weight ]
-payment-token = payment-method-or-wildcard "/" intent-or-wildcard
+payment-range  = payment-token *prefer-param [ weight ]
+payment-token  = payment-method-or-wildcard
+                 "/" intent-or-wildcard
+prefer-param   = OWS ";" OWS token "=" ( token / quoted-string )
 payment-method-or-wildcard = payment-method-id / "*"
-intent-or-wildcard = intent-token / "*"
+intent-or-wildcard         = intent-token / "*"
 
 ; Payment-Receipt header field value
 Payment-Receipt = base64url-nopad
--- a/specs/core/draft-httpauth-payment-00.md
+++ b/specs/core/draft-httpauth-payment-00.md
@@ -564,6 +564,46 @@ payment-method-or-wildcard = payment-method-id / "*"
 intent-or-wildcard         = intent-token / "*"
 ~~~
 
+### Preference Parameters {#preference-parameters}
+
+Two parameters are defined for use with `Accept-Payment`.
+Implementations MUST ignore unknown parameters.
+
+**`currency`**: The currency or asset the client can pay
+  with. The value uses the same format as the `currency`
+  field in the corresponding payment method's request
+  schema: ISO 4217 lowercase for fiat methods (e.g.,
+  `usd`), token contract address for on-chain methods
+  (e.g., `0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913`),
+  or a method-defined identifier (e.g., `sat` for
+  Lightning). When present, the server SHOULD only return
+  challenges whose `request` object specifies a matching
+  currency.
+
+**`network`**: The payment network or chain the client can
+  use. The format is defined by the payment method
+  specification: EIP-155 chain ID for EVM methods (e.g.,
+  `8453`), CAIP-2 identifier for Stellar (e.g.,
+  `stellar:pubnet`), a network name for Solana and
+  Lightning (e.g., `mainnet`), or omitted entirely for
+  methods without a network concept (e.g., Stripe, card).
+  When present, the server SHOULD only return challenges
+  whose `request` object targets a matching network.
+
+Both parameters are method-scoped: the valid values for a
+given method are defined by that method's specification,
+not by this document. Payment method specifications SHOULD
+document the accepted values for use in `Accept-Payment`.
+
+Parameters are compared using case-sensitive byte-for-byte
+string matching. A `payment-range` without a parameter
+matches challenges regardless of that dimension, preserving
+backward compatibility.
+
+Payment method specifications MAY define additional
+preference parameters beyond `currency` and `network`.
+Implementations MUST ignore unknown parameters.
+
 Examples:
 
 ~~~http
--- a/specs/core/draft-httpauth-payment-00.md
+++ b/specs/core/draft-httpauth-payment-00.md
@@ -854,6 +854,19 @@ Clients MUST NOT rely on the `description` parameter for payment
 verification. Malicious servers could provide a misleading description
 while the actual `request` payload requests a different amount.
 
+## Accept-Payment Information Disclosure
+
+The `Accept-Payment` header reveals which payment methods,
+currencies, and networks a client supports. This metadata
+is visible to the server and any intermediary that can
+observe request headers.
+
+Clients concerned about disclosing wallet capabilities
+SHOULD omit `Accept-Payment` and instead filter challenges
+locally after receiving the 402 response. Servers MUST NOT
+require `Accept-Payment` as a precondition for issuing
+challenges.
+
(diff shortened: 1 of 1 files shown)


brendanjryan:
Thanks for submitting! 

I think `currency` would make sense to elevate since it is a top-level field defined in https://paymentauth.org/draft-httpauth-payment-00.html , but I do not think that we should include `network`, since it is an implementation detail of an underlying payment method.

Given that this PR now only introduces `currency`, I don't think we need to strictly change the spec here. In practice, many clients are able to do FX / currency exchanges themselves so can easily work around a currency mismatch (unlike ap ayment method mismatch). We will keep this in mind and reconsider in the future if we see a strong need on deployed services

arditti:
@brendanjryan thank you! much appriciate your review! Agreed that `network` fitting cleanly in the layered architecture is genuinely tricky — happy to drop it from Accept-Payment if that's where we land. But I want to flag what's lost, because I don't think discovery fully covers it:

**The gap:** Without any `network` signal at negotiation time, an agent holding USDC on Base has no way to tell a server "don't bother issuing me a challenge for L1 or Arbitrum." The agent either:

1. Sends the request blind, receives N challenges, decodes every `request` blob to find a compatible chain, discards the rest (extra latency + wasted server work)
2. Gets only one challenge on a chain it doesn't support, and has to bridge (real gas cost, not viable for per-request micropayments — this is where the FX analogy breaks, since USDC-on-Base <-> USDC-on-L1 bridging isn't comparable to USD<->EUR FX)
3. Fails the request entirely

Discovery (`x-payment-info`) helps, but only for endpoints that publish an OpenAPI surface. Agent-to-agent and ad-hoc 402 flows are still blind.

**Architectural counter:** The "`network` is a method implementation detail" framing made sense when each method was single-chain. But the unified EVM method (#213) explicitly supports multiple chains in a single method — `chainId` is a first-class choice the client must make, not an implementation detail buried beneath it. Solana has mainnet/devnet/localnet. Stellar has pubnet/testnet. `network` has become a negotiation axis, not a hidden one.

If Accept-Payment really isn't the right home, what would be? Some options:

- **`x-payment-info` in discovery** — works for OpenAPI-described endpoints, not for 402s in the wild
- **A new top-level field in the charge intent** — elevates `network` to peer with `currency`, addresses the "top-level vs methodDetails" distinction you drew
- **A companion header like `Accept-Payment-Network`** — keeps Accept-Payment clean but adds proliferation

Genuinely open to any of these — the goal is just that agents can express chain preference *before* paying the cost of a 402 round trip. Happy to close #255 and regroup around whichever venue you think is right. Should I drop `network` from the current PR and open a separate discussion issue to scope out the network-negotiation problem?

PR #257, opened 2026-05-11
spec(evm-charge): add Radius SBC example

TJ-Frederick:
Adds a Radius example to the EVM charge spec alongside the existing MegaETH and Sei examples. Pure example addition — no normative changes.

Renames `draft-evm-charge-00.md` → `draft-evm-charge-01.md` per CONTRIBUTING.md (version bump in filename and frontmatter).

**Radius example details**

- EIP-155 chain ID `723487`
- SBC token at `0x33ad9e4BD16B69B5BFdED37D8B5D9fF9aba014Fb` (6 decimals)
- Permit2 at `0x000000000022D473030F116dDEE9F6B43aC78BA3`
- `credentialTypes: ["permit2", "transaction", "hash"]` with permit2 preferred

Radius is a stablecoin-native EVM platform (mainnet launched 2026-03-05) using Permit2 + EIP-2612 for x402 settlement. Docs: https://docs.radiustech.xyz

### DIFF
--- a/specs/methods/evm/draft-evm-charge-01.md
+++ b/specs/methods/evm/draft-evm-charge-01.md
@@ -1,8 +1,8 @@
 ---
 title: EVM Charge Intent for HTTP Payment Authentication
 abbrev: EVM Charge
-docname: draft-evm-charge-00
-version: 00
+docname: draft-evm-charge-01
+version: 01
 category: info
 ipr: noModificationTrust200902
 submissiontype: IETF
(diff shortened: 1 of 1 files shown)


brendanjryan:
thanks for the suggestion!

does the addition of this example change the scope or illustrate any new functionality other than radius can be used on MPP? 

If no, I would like to avoid merging as it sets a precedent where we should do this for _every_ EVM method in the formal spec

TJ-Frederick:
Got it, makes sense. Thanks!

PR #260, opened 2026-05-13
fix: align Stripe charge SPT terminology

brendanjryan:
## Summary

Updates the Stripe charge method to use Shared Payment Token terminology while preserving the existing Stripe SDK SPT creation example.

### DIFF
--- a/specs/methods/stripe/draft-stripe-charge-00.md
+++ b/specs/methods/stripe/draft-stripe-charge-00.md
@@ -199,15 +199,18 @@ base64url encoding, per {{I-D.httpauth-payment}}.
 The client fulfills this by creating an SPT using Stripe:
 
 ~~~ javascript
-const body = new URLSearchParams({
+const spt = await stripe.sharedPayment.issuedTokens.create({
   payment_method: 'pm_123',
-  'usage_limits[currency]': 'usd',
-  'usage_limits[max_amount]': '5000',
-  'usage_limits[expires_at]': expiresAt,
-  'seller_details[network_id]': 'profile_123'
+  usage_limits: {
+    currency: 'usd',
+    max_amount: 5000,
+    expires_at: Timestamp
+  },
+  seller_details: {
+    networkId: 'profile_123'
+  }
 });
-const res = await fetch('/create-spt', { method: 'POST', body });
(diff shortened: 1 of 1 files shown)


specs/methods/stripe/draft-stripe-charge-00.md
```diff
@@ -194,18 +199,15 @@ base64url encoding, per {{I-D.httpauth-payment}}.
 The client fulfills this by creating an SPT using Stripe:
 
 ~~~ javascript
-const spt = await stripe.sharedPayment.issuedTokens.create({
+const body = new URLSearchParams({
```
raubrey-stripe: are URLSearchParams a normal thing for a body?

specs/methods/stripe/draft-stripe-charge-00.md
```diff
@@ -194,18 +199,15 @@ base64url encoding, per {{I-D.httpauth-payment}}.
...
-    max_amount: 5000,
-    expires_at: Timestamp
-  },
-  seller_details: {
-    networkId: 'profile_123'
-  }
+  'usage_limits[currency]': 'usd',
+  'usage_limits[max_amount]': '5000',
+  'usage_limits[expires_at]': expiresAt,
+  'seller_details[network_id]': 'profile_123'
```
raubrey-stripe: noting that I think we can keep it as is because MPP things are using `networkId` but we have renamed this thing to `network_business_profile` as part of SPT public preview: https://docs.stripe.com/api/shared-payment/granted-token/object?api-version=2026-04-22.preview&rds=1

Not a problem for today

raubrey-stripe (commented):
I don't actually think I can approve, but looks good, thanks for the change

specs/methods/stripe/draft-stripe-charge-00.md
brendanjryan: thanks for catching -- this was from an unrelated change

specs/methods/stripe/draft-stripe-charge-00.md
brendanjryan: 👍

raubrey-stripe (approved):
ty ty

PR #261, opened 2026-05-18
fix: add missing hash credential type

Xeift:
Updated the specification to include four credential types instead of two, adding details about the `type="hash"` credential type.


brendanjryan:
Thanks for flagging! Merging as this is just a documentation bug and does not change the shape of the intent

brendanjryan (approved):

PR #262, opened 2026-05-18
feat: add ataCreationRequired directive to Solana charge intent

lgalabru:


brendanjryan (approved):
LGTM

chopmob-cloud:
@lgalabru — operational note from running Solana SPL settlement in production (AlgoVoi gateway, CCTP V2 Solana settle path live since 2026-05-04). The `ataCreationRequired` directive and its verification procedure match the gotcha class we hit in production directly.

**1. The silent-failure mode this prevents is real and easy to hit.** Our CCTP V2 path had a subtle bug where the receiver `mintRecipient` was initialised with the user's wallet pubkey instead of the canonical ATA PDA. The CCTP receive completed on-chain without obvious error, but the funds landed at an address from which the user's wallet could not spend (the wallet pubkey is not the SPL-token owner of any token account). The `ataCreationRequired` + canonical-PDA-MUST verification closes this entire class — implementers can no longer accidentally use the owner pubkey as a token-account address.

**2. The canonical-PDA requirement is the load-bearing line.** `findAssociatedTokenAddress(owner, mint, tokenProgram)` produces a deterministic PDA, but different wallet/SDK implementations have historically diverged on edge cases (token-2022 program ID vs legacy spl-token program ID, off-curve owner handling). The spec text's explicit `tokenProgram` field and "MUST be the canonical Associated Token Account PDA for the owner, mint, and token program" is exactly the right disambiguation. Worth keeping that text strict — any softening to "SHOULD" surfaces interop failures that look like silent value loss.

**3. The fee-payer authorization scoping is correctly bounded** by requiring explicit `ataCreationRequired: true` per split and forbidding fee-payer-funded creation for the top-level recipient or unmarked splits. The "If the top-level recipient's ATA does not exist, the server MUST NOT issue a challenge that requires creating it" line correctly pushes the existence check into challenge-generation rather than verification — implementers should expect a `getAccountInfo(ata_pda)` RPC call before issuing the 402. Worth noting non-normatively for implementers who may assume the verification path handles it.

PR reads ready for review from our side. Happy to bring more CCTP V2 / SPL operational data if useful for an implementation-notes section.

PR #266, opened 2026-05-25
fix: explicitly add externalId to PaymentWitness struct

### LINKED ISSUE #263: EVM Charge: clarify how to put externalId in witness struct
### Describe the change

In EVM Charge, Witness Data section:
https://github.com/tempoxyz/mpp-specs/blob/b3695145e03f95fa661eef174a6274184b8047af/specs/methods/evm/draft-evm-charge-00.md?plain=1#L449-L473
mentioned:
> When `externalId` is present in the challenge
request, the client MUST include it in the EIP-712
witness struct.

but did not explicitly say the order and the type of  `externalId` in witness struct. There's no `externalId` field in example witness struct.
https://github.com/tempoxyz/mpp-specs/blob/b3695145e03f95fa661eef174a6274184b8047af/specs/methods/evm/draft-evm-charge-00.md?plain=1#L461-L463

Since order matters in EIP-712 struct, and will affects the resulting hash, would be nice if we can explicitly specify the `externalId` field.
Not sure if this is correct, I think it should be:
~~~solidity
struct PaymentWitness {
    bytes32 challengeHash;
    string externalId;
}
~~~

### Additional context

_No response_

Xeift:
Adds `externalId` to `PaymentWitness` to match the existing EIP-712 witness requirement.
This does not change the struct already described in spec, it clarifies the schema in the `PaymentWitness` codeblock.

Fixes #263

### DIFF
--- a/specs/methods/evm/draft-evm-charge-00.md
+++ b/specs/methods/evm/draft-evm-charge-00.md
@@ -452,14 +452,16 @@ The Permit2 witness mechanism provides cryptographic
 binding between the payment authorization and the
 challenge. When `externalId` is present in the challenge
 request, the client MUST include it in the EIP-712
-witness struct. The server MUST verify the witness
-matches before submitting the transaction.
+witness struct. When `externalId` is absent from the challenge request,
+`PaymentWitness.externalId` is the empty string (`""`). The server MUST
+verify the witness matches before submitting the transaction.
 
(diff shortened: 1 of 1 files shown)


brendanjryan:
LGTM

brendanjryan (approved):

PR #272, opened 2026-06-01
Add usdc charge payment method

harshalbhangale-circle:
## Summary

This PR adds a `usdc` charge method for Payment Auth.

The method gives merchants a unified way to advertise USDC acceptance across supported chains and settlement paths. Clients still receive the concrete details they need for the selected path, including how to pay, verify, apply replay protection, and prove settlement.

In this draft, each supported payment path is modeled as a profile. The v00 profiles are direct EVM USDC, direct Solana USDC, Stacks USDCx, and Gateway Transfer.

## Motivation

USDC is available across multiple chains and settlement paths. Merchants may want to integrate at the asset level while still choosing the chains and paths they support.

The `usdc` method provides that common layer. A merchant can advertise `usdc` once and indicate the supported profiles. A client with EVM support can pay through the EVM profile. A client with Solana support can pay through the Solana profile. A client using Gateway can fund from an advertised source network while the merchant settles on its chosen destination network.

This keeps the merchant integration centered on USDC while giving clients the concrete path details they need for payment and verification.

## What This Adds

This draft defines the `usdc` charge method and four v00 profiles:

1. Direct EVM USDC charges.
2. Direct Solana USDC charges.
3. Stacks USDCx charges.
4. Gateway Transfer charges, where the merchant chooses the destination network and the payer funds from an advertised Gateway source network.

It also defines shared USDC behavior for asset identity, profile selection, replay protection, token controls, and receipts.

## Notes

USDC is a common payment asset across multiple supported environments. Giving it a method-level home makes discovery and integration clearer for merchants and clients.

Merchants get one method for USDC acceptance. Clients get the profile details needed for the path they support. Servers apply common USDC request and receipt rules, then use profile-specific verification for settlement.

Gateway Transfer extends the same model to cross-chain charges. A seller can choose where it wants to receive USDC, while the payer funds from another supported Gateway network.

This keeps the public method simple: `usdc` for the asset, profiles for the payment path.

## Verification

- `make lint`
- `make check`

## AI Assistance Disclosure

This draft was prepared and reviewed with significant AI assistance. I reviewed the generated content for technical correctness, RFC-style formatting, and the contribution requirements in `CONTRIBUTING.md`.

### DIFF
--- a/specs/methods/usdc/draft-usdc-charge-00.md
+++ b/specs/methods/usdc/draft-usdc-charge-00.md
@@ -46,6 +46,12 @@ normative:
   I-D.solana-charge:
     title: Solana Charge Intent for HTTP Payment Authentication
     target: https://paymentauth.org/draft-solana-charge-00.html
+  CAIP-2:
+    title: Chain Agnostic Improvement Proposal 2
+    target: https://chainagnostic.org/CAIPs/caip-2
+  CAIP-10:
+    title: Chain Agnostic Improvement Proposal 10
+    target: https://chainagnostic.org/CAIPs/caip-10
   SIP-005:
     title: Stacks Blocks, Transactions, and Accounts
     target: https://raw.githubusercontent.com/stacksgov/sips/main/sips/sip-005/sip-005-blocks-and-transactions.md
@@ -249,14 +255,18 @@ through Circle Gateway discovery and estimate APIs or a conforming SDK.
 
 ## Method Details
 
-`methodDetails.profile` selects the active `usdc` profile. Exactly one of
-`methodDetails.evm`, `methodDetails.solana`, `methodDetails.stacks`, or
-`methodDetails.gateway` MUST be present, and it MUST match
-`methodDetails.profile`.
+`methodDetails.type` selects the active `usdc` profile. Its value MUST
+be one of `evm`, `solana`, `stacks`, or `gateway`.
+
+The `methodDetails` object MUST include exactly one profile details
+object, and that object MUST use the same name as `methodDetails.type`.
+For example, when `methodDetails.type = "solana"`,
+`methodDetails.solana` MUST be present and `methodDetails.evm`,
+`methodDetails.stacks`, and `methodDetails.gateway` MUST be absent.
 
 | Field | Type | Required | Description |
 | --- | --- | --- | --- |
-| `profile` | string | REQUIRED | One of `evm`, `solana`, `stacks`, or `gateway`. |
+| `type` | string | REQUIRED | One of `evm`, `solana`, `stacks`, or `gateway`. |
 | `evm` | object | CONDITIONAL | EVM details. Required for EVM direct USDC charge. |
 | `solana` | object | CONDITIONAL | Solana details. Required for Solana direct USDC charge. |
 | `stacks` | object | CONDITIONAL | Stacks details. Required for Stacks USDCx charge. |
@@ -376,7 +386,11 @@ so the payer's source-network choice is constrained by
 | `destinationNetwork` | string | REQUIRED | CAIP-2 destination network where the merchant wants to receive USDC. |
 | `maxFee` | string | REQUIRED | Absolute upper bound on the signed Gateway authorization fee cap, in USDC base units. |
 | `maxFeeBps` | number | OPTIONAL | Additional ratio cap. When present, `authorizationFeeCap * 10000 <= request.amount * maxFeeBps`. |
-| `credentialTypes` | array | OPTIONAL | If present, MUST contain only `gateway-transfer` in v00. If absent, `gateway-transfer` is implied. |
+| `credentialTypes` | array | OPTIONAL | If present, MUST contain only `transfer` in v00. If absent, `transfer` is implied. |
+
+Gateway network fields use CAIP-2 {{CAIP-2}} identifiers. Gateway
+account identifiers use CAIP-10 {{CAIP-10}} account IDs in the form
+`<CAIP-2 network>:<account address>`.
 
 Circle Gateway maps source and destination networks to Gateway domains,
 token identifiers, wallet contracts, minter contracts, recipient setup
@@ -440,7 +454,7 @@ keccak256(UTF-8 bytes of JCS({
   "method": "usdc",
   "realm": <challenge.realm>,
   "intent": "charge",
-  "profile": "gateway",
+  "type": "gateway",
   "requestHash": <requestHash>,
   "sourceNetwork": <selected source network>,
   "destinationNetwork": <destination network>,
@@ -646,17 +660,18 @@ For Gateway Transfer, the credential object contains:
 | `payload` | object | REQUIRED | Gateway Transfer payload. |
 | `source` | string | REQUIRED | CAIP-10 source depositor account ID. |
 
-The Gateway Transfer `payload.type` MUST be `gateway-transfer`. The
-payload carries the selected source for this charge. It does not repeat
-the full `acceptedSources` list from the request.
+The Gateway Transfer `payload.type` MUST be `transfer`. This value is
+scoped to `methodDetails.type="gateway"`. The payload carries the
+selected source for this charge. It does not repeat the full
+`acceptedSources` list from the request.
 
 | Field | Type | Required | Description |
 | --- | --- | --- | --- |
-| `type` | string | REQUIRED | MUST be `gateway-transfer`. |
+| `type` | string | REQUIRED | MUST be `transfer`. |
 | `sourceNetwork` | string | REQUIRED | Selected CAIP-2 source network. MUST be one entry from `methodDetails.gateway.acceptedSources`. |
 | `destinationNetwork` | string | REQUIRED | CAIP-2 destination network. |
 | `maxFee` | string | REQUIRED | Maximum Gateway fee authorized by the payer, in USDC base units. |
-| `authorization` | object | REQUIRED | Circle Gateway signed authorization package, or SDK-produced versioned object, for the selected source network. It MUST expose the signed fields required by {{verification-procedure}} to the verifier. |
+| `authorization` | object | REQUIRED | Circle Gateway signed authorization package, or SDK-produced versioned object, for the selected source network. |
 
 The Gateway authorization package represents one Gateway transfer for
 the selected source network. Its internal encoding is Circle Gateway
@@ -664,7 +679,9 @@ versioned data, not a PaymentAuth extension point. When this document
 names Gateway fields such as source depositor, source signer,
 TransferSpec value, fee cap, destination recipient, recipient setup
 options, or salt, it refers to their Gateway semantics. The wire object
-MAY carry those values under versioned Gateway field names.
+MAY carry those values under versioned Gateway field names. The
+authorization object MUST expose the signed fields required by
+{{verification-procedure}} to the verifier.
 
 The surrounding PaymentAuth credential binds that Gateway package to
 `credential.challenge.id` and `credential.challenge.realm` by requiring
@@ -684,8 +701,8 @@ verification:
 2. Verify `method="usdc"` and `intent="charge"`.
 3. Verify `credential.challenge` matches the selected challenge.
 4. Verify the challenge has not expired.
-5. Verify `methodDetails.profile` is present and exactly one matching
-   profile object is present.
+5. Verify `methodDetails.type` is present, exactly one profile object
+   is present, and the present object key matches `methodDetails.type`.
 6. Verify `amount` is a positive integer in base units.
 7. Verify the selected token is a supported USDC asset form for the
    selected profile.
@@ -701,7 +718,7 @@ controls required by local policy.
 
 For EVM, servers then apply {{I-D.evm-charge}} verification with the
 USDC restrictions in {{evm-profile}}. Servers MUST verify
-`methodDetails.profile = "evm"`. For `payload.type="authorization"`,
+`methodDetails.type = "evm"`. For `payload.type="authorization"`,
 the EIP-3009 fields are carried directly in `payload` as defined by
 {{I-D.evm-charge}}. The server MUST verify `payload.to` and
 `request.recipient` identify the same 20-byte EVM address,
@@ -712,7 +729,7 @@ case-sensitive string equality.
 
 For Solana, servers then apply {{I-D.solana-charge}} verification with
 the USDC restrictions in {{solana-profile}}. Servers MUST verify
-`methodDetails.profile = "solana"`. Servers MUST also verify that
+`methodDetails.type = "solana"`. Servers MUST also verify that
 `payload.challengeBinding.type = "challenge-binding-v00"` and that
 `payload.challengeBinding.signature` authenticates the challenge-binding
 binding message under the same source key that authorizes the token
@@ -754,7 +771,7 @@ exposure exceed local policy.
 For Stacks, servers MUST verify:
 
 1. `payload.type = "transaction"`.
-2. `methodDetails.profile = "stacks"`.
+2. `methodDetails.type = "stacks"`.
 3. `source` uses `stacks:<chainId>:<standard-principal>` and the
    chain id matches `methodDetails.stacks.chainId`.
 4. `methodDetails.stacks.decimals = 6`.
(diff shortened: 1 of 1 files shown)


specs/methods/usdc/draft-usdc-charge-00.md
```diff
@@ -0,0 +1,1483 @@
...
+For Gateway Transfer, `currency` MUST be the case-sensitive literal
+`usdc`. This is a method-defined asset identifier, not an ISO currency
+code. The concrete source and destination token identities are resolved
+through Circle Gateway discovery and estimate APIs or a conforming SDK.
+
+## Method Details
+
+`methodDetails.profile` selects the active `usdc` profile. Exactly one of
+`methodDetails.evm`, `methodDetails.solana`, `methodDetails.stacks`, or
+`methodDetails.gateway` MUST be present, and it MUST match
```
brendanjryan: what does this mean? must match methoddetails.profile? 


won't the field always match the field?

specs/methods/usdc/draft-usdc-charge-00.md
```diff
@@ -0,0 +1,1483 @@
...
+`methodDetails.profile`.
+
+| Field | Type | Required | Description |
+| --- | --- | --- | --- |
+| `profile` | string | REQUIRED | One of `evm`, `solana`, `stacks`, or `gateway`. |
```
brendanjryan: not sure i love the name "profile" -- did we consider others like 

* kind
* type

specs/methods/usdc/draft-usdc-charge-00.md
```diff
@@ -0,0 +1,1483 @@
...
+| `decimals` | number | REQUIRED | MUST be `6`. |
+| `tokenProgram` | string | REQUIRED | MUST be the legacy SPL Token program ID in v00. |
+| `feePayer` | boolean | OPTIONAL | Whether the server pays network fees. |
+| `feePayerKey` | string | CONDITIONAL | Required when `feePayer=true`; absent otherwise. |
+
+v00 profiles native USDC on the legacy SPL Token program only. Servers
+MUST reject credentials whose `tokenProgram` is not:
+
+```text
+TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA
```
brendanjryan, line 333: 👍

specs/methods/usdc/draft-usdc-charge-00.md
```diff
@@ -0,0 +1,1483 @@
...
+The Gateway Transfer profile uses Circle Gateway to move native USDC
+from a payer-selected Gateway source network to the merchant's chosen
+destination network. The merchant receives USDC on `destinationNetwork`,
+so the payer's source-network choice is constrained by
+`acceptedSources` rather than by the merchant's settlement network.
+
+| Field | Type | Required | Description |
+| --- | --- | --- | --- |
+| `acceptedSources` | array | REQUIRED | CAIP-2 Gateway source networks the payer may use for this charge. |
+| `destinationNetwork` | string | REQUIRED | CAIP-2 destination network where the merchant wants to receive USDC. |
```
brendanjryan, line 411: does CAIP-2 cover networks like solana too? 

if so, why not use CAIP-2 globally for all names?

specs/methods/usdc/draft-usdc-charge-00.md
```diff
@@ -0,0 +1,1483 @@
...
+| `maxFee` | string | REQUIRED | Absolute upper bound on the signed Gateway authorization fee cap, in USDC base units. |
+| `maxFeeBps` | number | OPTIONAL | Additional ratio cap. When present, `authorizationFeeCap * 10000 <= request.amount * maxFeeBps`. |
+| `credentialTypes` | array | OPTIONAL | If present, MUST contain only `gateway-transfer` in v00. If absent, `gateway-transfer` is implied. |
```
brendanjryan: since this is aleady under gateway, can this just be `transfer` ?

specs/methods/usdc/draft-usdc-charge-00.md
```diff
@@ -0,0 +1,1483 @@
...
+| Field | Type | Required | Description |
+| --- | --- | --- | --- |
+| `type` | string | REQUIRED | MUST be `challenge-binding-v00`. |
+| `signature` | string | REQUIRED | Base64-encoded 64-byte Ed25519 signature. |
+
+The client MUST sign the following JCS-canonicalized binding message:
+
+```text
+{
+  "type": "challenge-binding-v00",
```
brendanjryan: why do we need to do this?

specs/methods/usdc/draft-usdc-charge-00.md
```diff
@@ -0,0 +1,1483 @@
...
+The Stacks transaction MUST call the SIP-010 `transfer` function with
+`amount`, `sender`, `recipient`, and optional memo arguments. It MUST
+include a post-condition that pins a `SentEq` transfer of
+`request.amount` for the advertised USDCx asset.
+
+`payload.challengeBinding.type` MUST be `challenge-binding-v00`.
+`payload.challengeBinding.signature` MUST be a base64-encoded 65-byte
+recoverable secp256k1 signature in `r || s || v` order, with `v` in
+`{0x00, 0x01}`. This document defines the SIP-018 domain tuple for the
+challenge-binding message. The `USDCPaymentAuth` name is a
```
brendanjryan: similar to above -- why is this strictly needed?

specs/methods/usdc/draft-usdc-charge-00.md
```diff
@@ -0,0 +1,1483 @@
...
+
+| Field | Type | Required | Description |
+| --- | --- | --- | --- |
+| `method` | string | REQUIRED | MUST be `usdc`. |
+| `intent` | string | REQUIRED | MUST be `charge`. |
+| `challengeId` | string | REQUIRED | Original challenge ID. |
+| `reference` | string | REQUIRED | Transaction hash, signature, or transaction ID. |
+| `referenceType` | string | REQUIRED | `tx_hash`, `tx_signature`, or `tx_id`. |
+| `status` | string | REQUIRED | `success` only after the selected profile has completed settlement. |
+| `timestamp` | string | REQUIRED | RFC3339 settlement time. |
```
brendanjryan: why do we need to add all of these fields to the receipt? are all of these strictly necessary? 

this is a large addition to the existing types and would make it harder to write an SDK

specs/methods/usdc/draft-usdc-charge-00.md
harshalbhangale-circle: Yes, that wording was awkward. The intent was to say that `methodDetails.type` is the discriminator, and exactly one matching profile object should be present. I updated this to use `type` instead of `profile`

specs/methods/usdc/draft-usdc-charge-00.md
harshalbhangale-circle: changed it to type

specs/methods/usdc/draft-usdc-charge-00.md
harshalbhangale-circle: CAIP-2 does cover Solana. I kept CAIP-2 for Gateway source and destination network fields because Gateway can route across chain families. I did not switch the direct EVM and Solana profiles to CAIP-2 globally because those profiles inherit the identifier formats from the existing EVM and Solana PaymentAuth specs. But I dont have a strong preference

specs/methods/usdc/draft-usdc-charge-00.md
harshalbhangale-circle: Thanks, updated

specs/methods/usdc/draft-usdc-charge-00.md
harshalbhangale-circle: That makes sense. Reduced the receipt shape substantially. This should be easier for SDKs while still preserving enough settlement context for the selected USDC profile.

specs/methods/usdc/draft-usdc-charge-00.md
harshalbhangale-circle: The Solana transaction signature proves that the payer authorized the transaction bytes, but it does not itself sign the PaymentAuth challenge. The detached binding adds a same-key signature over the exact challenge, source account, and transaction bytes digest. That gives the profile challenge-bound payer authorization similar in spirit to EVM, while keeping the challenge id offchain.

The tradeoff is extra client and SDK complexity. If we want v00 to stay closer to the existing Solana pull-mode model, I’m fine removing the detached binding for Solana and using the base verification semantics. For Stacks, where there is no base method to inherit, we can define the same server-broadcast, verify-terms, consume-transaction-id model without the second signature.

specs/methods/usdc/draft-usdc-charge-00.md
```diff
@@ -0,0 +1,1433 @@
...
+`charge`, another intent, or a different request with the same amount and
+recipient.
+
+For Solana transaction credentials, the credential object contains:
+
+| Field | Type | Required | Description |
+| --- | --- | --- | --- |
+| `challenge` | object | REQUIRED | Echo of the server challenge. |
+| `payload` | object | REQUIRED | Solana payment payload. |
+| `source` | string | REQUIRED | CAIP-10 account ID using `solana:<genesis-hash-prefix>:<pubkey>`. |
```
brendanjryan: **`source` encoding contradicts core/evm/solana.**

This draft requires bare CAIP-10 for `source` across all profiles (here `solana:<genesis>:<pubkey>`, and `eip155:5042002:0x…` in the EVM example at line 1121). But:

- **core** recommends DID format (`source` = "Payer identifier (RECOMMENDED: DID format per [W3C-DID])")
- **evm-charge** §5.1: `source` SHOULD be `did:pkh:eip155:…`
- **solana-charge**: `source` MAY be a base58 pubkey or a DID

So the EVM example `"source":"eip155:…"` directly contradicts evm-charge's `did:pkh:eip155:…`. This draft also makes `source` **REQUIRED**, whereas all three base specs make it **OPTIONAL**.

Please either adopt the DID/`did:pkh` convention used by the rest of the family, or get core/evm/solana updated to bless CAIP-10 — and reconcile the required-vs-optional difference.

specs/methods/usdc/draft-usdc-charge-00.md
```diff
@@ -0,0 +1,1433 @@
...
+  "realm": <challenge.realm>,
+  "intent": "charge",
+  "requestHash": <requestHash>
+}))
+```
+
+`requestHash` is `keccak256` of the UTF-8 bytes of the exact
+JCS-canonicalized request JSON before base64url encoding.
+
+This is stricter than the base EVM charge profile. The base EVM profile
```
brendanjryan: **EVM authorization nonce derivation conflicts with evm-charge**

evm-charge §5.3.1 defines `nonce = keccak256(abi.encodePacked(challenge.id, challenge.realm))`. This draft (line 516) defines `nonce = keccak256(JCS({id, method, realm, intent, requestHash}))`. These are not interchangeable.

The text just above says the EVM profile "inherit[s] {{I-D.evm-charge}} authorization payloads," but a generic evm-charge verifier will compute a different expected nonce and reject these credentials. The stricter binding is reasonable, but please reword to state explicitly that the nonce binding is **overridden** (not inherited) and that `usdc` EVM credentials are not interchangeable with `method="evm"`.

specs/methods/usdc/draft-usdc-charge-00.md
```diff
@@ -0,0 +1,1433 @@
...
+transaction verification.
+
+For Gateway Transfer, `currency` MUST be the case-sensitive literal
+`usdc`. This is a method-defined asset identifier, not an ISO currency
+code. The concrete source and destination token identities are resolved
+through Circle Gateway discovery and estimate APIs or a conforming SDK.
+
+## Method Details
+
+`methodDetails.type` selects the active `usdc` profile. Its value MUST
```
brendanjryan, line 255: **`methodDetails` shape diverges from evm-charge/solana-charge**

Base evm-charge and solana-charge place method fields flat in `methodDetails` (`methodDetails.chainId`, `methodDetails.network`). This draft introduces a `methodDetails.type` discriminator plus a nested profile object (`methodDetails.evm.chainId`). 

I think this is a sensible design for a method multiplexing four profiles, but it means the EVM/Solana request envelope is **not literally inherited** — a base-spec `methodDetails` parser won't work here. 

Please call this divergence out explicitly so implementers don't assume drop-in reuse.

specs/methods/usdc/draft-usdc-charge-00.md
```diff
@@ -0,0 +1,1433 @@
...
+
+| Field | Type | Description |
+| --- | --- | --- |
+| `method` | string | MUST be `usdc`. |
+| `type` | string | Selected USDC profile: `evm`, `solana`, `stacks`, or `gateway`. |
+| `challengeId` | string | Original challenge ID. |
+| `reference` | string | Final settlement reference. |
+| `status` | string | MUST be `success` only after the selected profile has completed settlement. |
+| `timestamp` | string | RFC3339 settlement time. |
+| `network` | string | Settlement network identifier. |
```
brendanjryan: **Receipt settlement-locator inconsistent with sibling specs**

This receipt uses `network` (CAIP-2 string) for every profile. But evm-charge's receipt uses `chainId` (number), and solana-charge's receipt has no network field at all. Cross-method receipt consumers will see a different locator field depending on method.

Consider either keeping `chainId` (number) for the EVM profile to match evm-charge, or explicitly documenting the deliberate switch to a unified CAIP-2 `network` field and noting it differs from the base EVM/Solana receipts.

brendanjryan (commented):
Left some final reviews! Overall i'm not fully convinced for the need for CAIP prefixes as all of the methods are scoped to a specific chain as is. 

In other methods (solana, evm etc..) we just use the chaindId field to discriminate, as the prefix is implicit in the method (e.g. all solana chain ids are implicitly under solana)

specs/methods/usdc/draft-usdc-charge-00.md
harshalbhangale-circle: Agreed. There's now a note right after the methodDetails table

specs/methods/usdc/draft-usdc-charge-00.md
harshalbhangale-circle: Reworded. It now says the nonce derivation overrides (does not inherit) the base evm-charge derivation

specs/methods/usdc/draft-usdc-charge-00.md
harshalbhangale-circle: Kept the unified field but documented it. network is now typed CAIP-2 in the table, with a note that it's a deliberate, method-wide settlement locator that differs from the base EVM receipt (numeric chainId) and the base Solana receipt (no network field). A usdc receipt is asset-level, so a consumer reads one locator regardless of profile, and a Gateway destination may not be EVM, so a numeric chainId wouldn't work there anyway.

specs/methods/usdc/draft-usdc-charge-00.md
harshalbhangale-circle: Decided to drop it for v00. The Solana profile now just inherits base pull-mode verification from solana-charge, no detached binding.

specs/methods/usdc/draft-usdc-charge-00.md
harshalbhangale-circle: removed for v00. Stacks now defines plain server-broadcast verification: check the SIP-010 transfer, its post-condition, and the origin signature, then bind to the challenge through challenge consumption and transaction-id replay. No second SIP-018 signature

specs/methods/usdc/draft-usdc-charge-00.md
harshalbhangale-circle: Thanks for the catch. Fixed it. Reconciled per profile, aligning the direct profiles with their base specs:

  - EVM: source is now OPTIONAL with did:pkh:eip155:… RECOMMENDED, matching evm-charge §5.1. The line-1121 example now reads did:pkh:eip155:5042002:0x….
  - Solana: source OPTIONAL, may be a base58 pubkey or a DID (did:pkh:solana:… recommended), matching solana-charge. Example updated to did:pkh:….
  - Gateway Transfer: source stays REQUIRED and bare CAIP-10. It's the source depositor account, and Gateway needs it named to build and verify the burn/transfer authorization, so it can't
  be optional here. There's no base method for it to contradict.
  - Stacks: source stays REQUIRED and CAIP-10 (stacks:<chainId>:<principal>). No base MPP method to inherit a DID convention from, and the origin signature recovers to this principal
  during verification.

specs/methods/usdc/draft-usdc-charge-00.md
brendanjryan: I see -- this makes sense. i'm ok keeping with this

brendanjryan (approved):
LGTM -- thanks for all of the revisions on this!

PR #284, opened 2026-06-18
Add Near Intents charge intent

IkerAlus:
## Summary

Adds the `nearintents` payment method implementing the existing `charge` intent. It settles one-time, cross-chain payments through the NEAR Intents 1Click Swap API: the client pays a source asset on any supported origin chain and the merchant receives an exact destination asset on any supported destination chain, with the NEAR Intents solver network executing the swap in between.

## Notes for reviewers

- **Push mode only** (`type="hash"`): The source asset must be deposited before settlement, so there is no server-submitted/pull flow; the client always pays
  its own origin-chain network fee.
- **Cross-chain field mapping**: The standard `amount`/`currency`/`recipient` describe the client's origin-chain deposit, with `recipient` = the 1Click deposit address; the merchant's destination leg lives in `methodDetails`. This keeps push-mode verification ("confirm a transfer of `amount` `currency` to `recipient`") directly applicable.
- **Challenge binding**: 1Click issues a unique deposit address per quote, so `recipient` is challenge-specific and bound into the challenge `id` — a deposit
  cannot be replayed against a different challenge.
- **Trust model**: funds are held by the settlement backend during the swap, with an automatic refund to `refundTo` on any non-success terminal state.

Registers a new method against the already-standardized `charge` intent; it does not define a new intent.

## Checklist

- [ ] `make check` passes
- [x] `make lint` passes
- [x] New method follows the method template, added under `specs/methods/nearintents/`
- [x] References are stable (no hardcoded section pointers)

## AI assistance

Drafting and formatting were done with AI assistance. All content has been reviewed by the author for RFC compliance and technical correctness.


specs/methods/nearintents/draft-nearintents-charge-00.md
```diff
@@ -0,0 +1,883 @@
...
+## Asset Identifiers {#asset-identifiers}
+
+The format of `currency` (and of `methodDetails.destinationAsset`)
+depends on the chain:
+
+| Chain type                         | Identifier format                | Example                                                            |
+| ---------------------------------- | -------------------------------- | ------------------------------------------------------------------ |
+| EVM chains                         | ERC-20 contract address          | `0xaf88d065e77c8cC2239327C5EDb3A432268e5831`                       |
+| Solana                             | SPL token mint address           | `EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v`                      |
+| NEAR                               | NEP-141 contract account ID      | `wrap.near`                                                        |
```
brendanjryan: **Inconsistent asset-identifier format.** 

This table says NEAR assets are a bare "NEP-141 contract account ID" (`wrap.near`), but every example expresses `destinationAsset` in the CAIP-19-style prefixed form `nep141:17208628f84f...` (lines 324, 753, 839).

Can we standardize this?

specs/methods/nearintents/draft-nearintents-charge-00.md
```diff
@@ -0,0 +1,883 @@
...
+base64url encoding, per {{I-D.httpauth-payment}}.
+
+This specification implements the shared request fields defined in
+{{I-D.payment-intent-charge}}.
+
+## Shared Fields
+
+| Field         | Type   | Presence | Description                                                                                          |
+| ------------- | ------ | -------- | ---------------------------------------------------------------------------------------------------- |
+| `amount`      | string | REQUIRED | Deposit amount the client sends on the origin chain, in base units of `currency` (stringified integer). Corresponds to the 1Click quote `maxAmountIn`. |
```
brendanjryan: **`amount` vs `minAmountIn` mismatch.** The cross-chain model and this row frame the invariant as "confirm a transfer of `amount` `currency` to `recipient`", and `amount` is defined as the deposit the client sends (= quote `maxAmountIn`). But verification step 3 actually accepts a transfer of `>= methodDetails.minAmountIn`, and `payment-insufficient` is keyed off `minAmountIn` — so acceptance is bound to `minAmountIn`, not `amount`.

Can you verify this?

specs/methods/nearintents/draft-nearintents-charge-00.md
```diff
@@ -0,0 +1,883 @@
...
+# Introduction
+
+HTTP Payment Authentication {{I-D.httpauth-payment}} defines a
+challenge-response mechanism that gates access to resources behind
+payments. This document registers the "charge" intent for the
+"nearintents" payment method.
+
+NEAR Intents is an intent-based settlement system in which a network of
+competing solvers fulfills user-expressed swap intents. The 1Click Swap
+API {{ONECLICK-API}} is a hosted interface to this system: a caller
```
brendanjryan, line 102: is 1click swap API relevant here? I would prefer to remove from the generic spec if not

specs/methods/nearintents/draft-nearintents-charge-00.md
```diff
@@ -0,0 +1,883 @@
...
+  |                        |--------------------|                  |
+  |                        | (8) Submit + await swap finality      |
+  |                        |-------------------------------------->|
+  |                        |   status: SUCCESS (merchant paid)     |
+  |                        |<--------------------------------------|
+  | (9) 200 OK + Receipt   |                    |                  |
+  |<-----------------------|                    |                  |
+~~~
+
+Unlike methods that offer a server-submitted ("pull") flow for fee
```
brendanjryan, line 160: 👍

specs/methods/nearintents/draft-nearintents-charge-00.md
```diff
@@ -0,0 +1,883 @@
...
+leg** is carried in `methodDetails`:
+
+- `recipient` is the 1Click **deposit address** on the origin chain. It is
+  the payee of the client's on-chain transfer and the address whose
+  receipt of funds the server verifies from the transaction hash. It is
+  controlled by the NEAR Intents settlement system for the duration of the
+  swap, not by the merchant.
+- `currency` and `amount` describe the **source asset** and the **deposit
+  amount** the client sends to `recipient`.
+- `methodDetails.network` is the **origin chain** where `recipient` lives
```
brendanjryan: to make this clearer, why not have this be originNetwork?

brendanjryan:
Thanks for submitting! Overall spec looks strong and consistent -- left some minor comments around documentation consistencies and names

specs/methods/nearintents/draft-nearintents-charge-00.md
IkerAlus: Actually the 1 click swap API is the service being called by the server to get a quote, check status etc. We can omit the specific service being queried but the implementation will be based in 1 click swap API endpoints in practice.

specs/methods/nearintents/draft-nearintents-charge-00.md
IkerAlus: Good catch, fixed by standardizing on CAIP-19 for both legs. The table, the currency/destinationAsset field rows, and every example now use the prefixed CAIP-19 form for the source and destination. CAIP-19 and CAIP-2 are now normative references. I also added two consistency rules: the chain component of currency MUST equal `methodDetails.originNetwork` and the chain component of destinationAsset MUST equal `methodDetails.destinationNetwork`

specs/methods/nearintents/draft-nearintents-charge-00.md
IkerAlus: good point, I fixed this as: The push-mode invariant no longer says "transfer of amount"; it now states the verifier's required threshold is `methodDetails.minAmountIn` (the acceptance floor), and defines amount as the deposit the client is asked to send.

specs/methods/nearintents/draft-nearintents-charge-00.md
IkerAlus: Done, renamed `methodDetails.network` to `methodDetails.originNetwork` everywhere.

IkerAlus:
hi @brendanjryan  thank you very much for the prompt review,  I addressed all the comments, the spec is ready for a new round of reviews.

Note: I also updated the spec as per the "Hash is consumed before settlement complete" point that may have been deleted after your review. I do think it was a valid point to improve here: I took the "defer consumption" path you suggested and documented the recovery. Verification now only claims the hash as in-flight (enough to satisfy the concurrency requirement); permanent consumption happens in settlement on any terminal state. I added a Client Recovery subsection for this too. Let me know if this makes the refund process clearer or it is better to simply leave it as it was.

specs/methods/nearintents/draft-nearintents-charge-00.md
brendanjryan: make sense! I would just rather leave specific service names (which will change over time) out of the spec

brendanjryan (approved):
LGTM! Thanks for the revisions

PR #285, opened 2026-06-19
docs: require non-empty challenge ids

EfeDurmaz16:
## Summary
- Clarifies that Payment challenge `id` is required and must be non-empty after auth-param parsing and quoted-string unescaping.
- States that servers must not emit missing or empty ids and clients/parsers must reject them.
- Leaves the generic auth-param ABNF intact and records the non-empty constraint in prose.

This follows the direction in https://github.com/tempoxyz/mpp-tools/issues/19#issuecomment-4752175102.

## Verification
- Attempted `./scripts/check.sh`; blocked locally by missing `kramdown-rfc` and `xml2rfc`.
- Attempted `make check`; blocked because Docker/OrbStack socket was unavailable.

### DIFF
--- a/specs/core/draft-httpauth-payment-00.md
+++ b/specs/core/draft-httpauth-payment-00.md
@@ -240,9 +240,13 @@ auth-param      = token BWS "=" BWS ( token / quoted-string )
 
 ### Required Parameters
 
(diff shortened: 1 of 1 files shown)


brendanjryan (approved):
LGTM

PR #296, opened 2026-07-13
feat(solana/sessions): operational costs reduction

lgalabru:
We're about to launch a library backing up the solana session intent.
During our end to end testing, we noticed that operational costs could be cut by an order of magnitude, by revisiting the tomb-stoning mechanics.
This PR is also tightening some semantics (`Finalize` -> `Seal`).


brendanjryan (approved):

brendanjryan:
LGTM! Excited to see the SDK

PR #299, opened 2026-07-15
Clarify Near Intents refund and settlement recovery

mikedotexe:
## Summary

Revise the Near Intents charge draft from -00 to -01 and align its normative safety/recovery behavior with implementation hardening in IkerAlus/mpp-nearintents#9.

- resolve and challenge-bind the refund destination before requesting a wet quote; clients validate it before paying
- match the presented origin transaction hash before consuming it or retiring a terminal quote
- separate terminal backend outcomes from transport and polling timeouts
- document the extra unauthenticated retry when a framework has already computed the terminal 402 challenge
- require chain-canonical transaction identifier comparison rather than blind lowercasing
- separate payment replay protection from application response idempotency

No wire fields or method identifiers change.

## Validation

- frontmatter and external-section-reference lint
- 25 repository tests
- full XML, HTML, TXT, and PDF generation
- rfclint and git diff check

Docker Desktop was unavailable, so the same local generation and validation scripts ran with the repository dependencies installed into temporary Ruby and Python environments.

## AI assistance disclosure

Codex was used to help compare the implementation and specification, draft the edits, and run the validation workflow. I reviewed the resulting normative text and am responsible for its correctness.

### DIFF
--- a/specs/methods/nearintents/draft-nearintents-charge-01.md
+++ b/specs/methods/nearintents/draft-nearintents-charge-01.md
@@ -1,8 +1,8 @@
 ---
 title: "NEAR Intents Charge Intent for HTTP Payment Authentication"
 abbrev: "NEAR Intents Charge"
-docname: draft-nearintents-charge-00
-version: 00
+docname: draft-nearintents-charge-01
+version: "01"
 category: info
 ipr: noModificationTrust200902
 submissiontype: independent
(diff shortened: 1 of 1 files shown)


specs/methods/nearintents/draft-nearintents-charge-01.md
```diff
@@ -315,6 +315,15 @@ If `methodDetails.credentialTypes` is omitted, servers MUST accept
 `"hash"`. Servers MUST NOT advertise credential types other than
 `"hash"` for this method.
 
+The server MUST determine `methodDetails.refundTo` before requesting
+the wet 1Click quote. It SHOULD resolve an origin-chain address
```
brendanjryan: what is a wet quote?

brendanjryan:
generally LGTM -- one terminology question

specs/methods/nearintents/draft-nearintents-charge-01.md
mikedotexe: Good catch. By “wet quote” I meant the non-dry/executable quote from the NEAR Intents 1Click Swap API: the quote request made after `refundTo` is known that allocates the single-use deposit address and can actually be paid. That wording is too ecosystem-local for this draft, so I’ll replace it with something like “executable 1Click quote” or “the 1Click quote that returns the deposit address.”

mikedotexe: Patched in 0923aee. The draft now says “the executable 1Click quote from the NEAR Intents 1Click Swap API: the quote that returns the single-use deposit address and can actually be paid,” so it avoids the ecosystem-local “wet quote” shorthand.

brendanjryan:
LGTM

brendanjryan (approved):

PR #270, opened 2026-05-28
Solana subscription intent specification 

lgalabru:

PR #310, opened 2026-07-31
feat(solana/subscription): new intent

lgalabru:
## Summary
- restore the Solana subscription intent draft from #270 on current `main`
- align Solana request and receipt fields with the shared subscription intent
- constrain renewals to exactly one full `amount` transfer per billing period
- clarify the on-chain plan, puller, program ID, expiry, and cancellation semantics

## Test Plan
- `make check`
- targeted frontmatter lint for `draft-solana-subscription-00.md`
- `git diff --check`

## Review Feedback Addressed
- preserve shared `externalId` semantics and move the Plan PDA to `methodDetails.planAddress`
- keep `subscriptionId` opaque base64url and expose `subscriptionDelegation` separately
- make `subscriptionExpires` activation validation conditional
- require exact full-amount renewals and remove duplicated renewal prose
- make `description` optional and include it in request examples
- require `methodDetails.programId`
- document that `puller` matches the subscriptions program vocabulary
- use a numeric `periodIndex` and rename receipt timestamps to `periodStart` and `periodEnd`

This replaces #270, whose deleted head branch GitHub would not reattach after restoration.

## AI Assistance
Codex assisted with reconciling the review feedback, editing the draft, and running validation.

### DIFF
--- a/specs/methods/solana/draft-solana-subscription-00.md
+++ b/specs/methods/solana/draft-solana-subscription-00.md
@@ -272,8 +272,8 @@ Solana uses the shared `amount`, `currency`, `periodUnit`,
 `periodCount`, `subscriptionExpires`, `recipient`, `description`, and
 `externalId` fields from {{I-D.payment-intent-subscription}}, with
 their meanings preserved. The Solana profile elevates `recipient`,
-`externalId`, `subscriptionExpires`, and `description` from OPTIONAL
-to REQUIRED, and constrains the values that `periodUnit` may take.
+`externalId`, and `description` from OPTIONAL to REQUIRED, and
+constrains the values that `periodUnit` may take.
 
 ### Required Fields
 
@@ -285,9 +285,14 @@ to REQUIRED, and constrains the values that `periodUnit` may take.
 | `periodCount` | string | Positive integer count of `periodUnit` values per billing period |
 | `recipient` | string | Recipient address authorized for subscription charges. The activation transaction MUST bind the destination at sign time |
 | `externalId` | string | Base58 address of the on-chain `Plan` |
-| `subscriptionExpires` | string | Subscription expiry timestamp in {{RFC3339}} format |
 | `description` | string | Human-readable subscription description |
 | `methodDetails` | object | Solana-specific extension data (see {{method-details}}) |
+
+### Optional Fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `subscriptionExpires` | string | Subscription expiry timestamp in {{RFC3339}} format. When omitted, the subscription has no HTTP-layer maximum lifetime |
 
 The `amount` value MUST be a string representation of a positive
 integer in base 10 with no sign, decimal point, exponent, or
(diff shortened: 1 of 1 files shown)


specs/methods/solana/draft-solana-subscription-00.md
```diff
@@ -0,0 +1,1225 @@
...
+
+### Required Fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `amount` | string | Fixed payment amount per billing period in base units |
+| `currency` | string | SPL Token or Token-2022 mint address (see {{currency-formats}}) |
+| `periodUnit` | string | Billing period unit. The value MUST be `day` or `week` |
+| `periodCount` | string | Positive integer count of `periodUnit` values per billing period |
+| `recipient` | string | Recipient address authorized for subscription charges. The activation transaction MUST bind the destination at sign time |
```
brendanjryan: `recipient` is elevated to REQUIRED here and the row says the activation transaction "MUST bind the destination at sign time", but nothing downstream actually ties the two together. The activation check in Credential Processing verifies program ID, plan, token program, mint, puller, `plan.destinations`, and amount, and the renewal checks only bind `methodDetails.puller`. Plan Scope Minimization then describes `plan.destinations` as a whitelist that can be empty (open). So a server can advertise `recipient: A` in the challenge, and the `transfer_subscription` at activation (and every renewal) can legally land in any `destination_ata` the puller picks, with the client only told to "verify `recipient` is controlled by the expected party".

Can we either make this a real binding, i.e. servers MUST verify `owner(destination_ata) == recipient` on activation and on every renewal (and reject if `plan.destinations` is non-empty and does not include it), or drop `recipient` back to OPTIONAL/informational as in the base intent? Right now it reads as a guarantee the protocol does not provide.

specs/methods/solana/draft-solana-subscription-00.md
```diff
@@ -0,0 +1,1225 @@
...
+|-------|------|----------|-------------|
+| `challenge` | object | REQUIRED | Echo of the challenge from the server |
+| `payload` | object | REQUIRED | Solana-specific activation or access payload |
+| `source` | string | OPTIONAL | Subscriber identifier (e.g., `did:pkh:solana:...`) |
+
+Subscriptions on Solana MUST use one of two activation-payload types.
+In pull mode (`type="transaction"`), the client signs the activation
+transaction and submits the serialized bytes; the server co-signs as
+fee payer if configured and broadcasts. In push mode
+(`type="signature"`), the client broadcasts the activation
```
brendanjryan: I don't think `type="signature"` is consistent with the rest of the activation model. `methodDetails.puller` is REQUIRED to be the server's puller key, the example activation has `Instruction 4: subscriptions.transfer_subscription, signer: puller` with "puller (server, added at co-sign)", and the flow diagram shows the server co-signing and broadcasting. If the client broadcasts the activation itself, the server never signs it, so the first-period `transfer_subscription` can't be in that transaction. Either the activation in this mode is `subscribe` only and the first pull happens out of band (in which case the draft needs to say when it happens and what the server verifies before granting access), or there's a program path where the puller signature isn't needed, which isn't described anywhere.

Can we spell out how the first-period transfer is authorized under `type="signature"`, or remove that mode from this draft until it's specified? The `feePayer: true` rejection rule in the next paragraph hints at the problem but doesn't resolve it.

lgalabru:
Thanks @brendanjryan!
Feedback addressed in https://github.com/tempoxyz/mpp-specs/pull/310/commits/5c4a528ce113815a2fb0b79801bd62bb2471a0e8

PR #313, opened 2026-08-08
fix(evm-charge): remove redundant externalId from Permit2 witness

### LINKED ISSUE #287: EVM Charge: Remove externalId in Permit2 Payment Witness Struct
### Describe the change

I suggested adding externalId to the Permit2 witness struct in #263 before, since the spec said the client MUST include externalId in the witness struct.

However, I actually didn't read the full payment-00 and charge-00 specs, I only read evm-charge-00 at that time. Now I've dived deep into these 3 specs, and I found some problems about the data to put in the Permit2 payment witness struct. Below are the suggested change and thought.

### 1. Completely remove externalId in Permit2 payment witness struct
Current witness struct is [defined](https://github.com/tempoxyz/mpp-specs/blob/ddeb95f28e90085f49ad0413c441b08e0bcfe4a8/specs/methods/evm/draft-evm-charge-00.md?plain=1#L461-L475) as below:
~~~solidity
struct PaymentWitness {
    bytes32 challengeHash;
    string externalId;
}

challengeHash = keccak256(abi.encodePacked(
    challenge.id,
    challenge.realm
))
~~~

I believe we can remove the externalId:
~~~solidity
struct PaymentWitness {
    bytes32 challengeHash;
    // 👈 remove the externalId
}

challengeHash = keccak256(abi.encodePacked(
    challenge.id,
    challenge.realm
))
~~~

This does not remove the challenge binding. Reason:
- We have `challengeHash` in the `PaymentWitness` struct ([EVM Charge](https://github.com/tempoxyz/mpp-specs/blob/ddeb95f28e90085f49ad0413c441b08e0bcfe4a8/specs/methods/evm/draft-evm-charge-00.md?plain=1#L461-L466))
- `challengeHash` is computed using `challenge.id` ([EVM Charge](https://github.com/tempoxyz/mpp-specs/blob/ddeb95f28e90085f49ad0413c441b08e0bcfe4a8/specs/methods/evm/draft-evm-charge-00.md?plain=1#L468-L475))
- `challenge.id` is computed using `base64url(HMAC-SHA256(secretKey, realm|method|intent|serialize(request👈)|expires|digest|opaque))` ([Core](https://github.com/tempoxyz/mpp-specs/blob/ddeb95f28e90085f49ad0413c441b08e0bcfe4a8/specs/core/draft-httpauth-payment-00.md?plain=1#L311-L326))
- `challenge.request` contains `amount`, `currency`, `recipient`, `description`, `externalId`👈 ([EVM Charge](https://github.com/tempoxyz/mpp-specs/blob/ddeb95f28e90085f49ad0413c441b08e0bcfe4a8/specs/methods/evm/draft-evm-charge-00.md?plain=1#L251-L266))

That is,
- `externalId` is already bound by `challenge.id` through `request`
- Server will verify if `credential.challenge.id === base64url(HMAC-SHA256(secretKey, realm|method|intent|serialize(credential.challenge.request)|expires|digest|opaque))`
- If a malicious client modifies the `externalId`, then the `request` no longer matches `challenge.id`, thus challenge binding verification fails

### 2. Adjust the field for challenge binding
Currently EIP-3009 does not put externalId in challenge binding (`nonce` field):
~~~solidity
nonce = keccak256(abi.encodePacked(
    challenge.id,
    challenge.realm
))
~~~
I think the challenge binding should align with Permit2. So if we eventually decide to keep the `externalId`, then challenge binding for the `authorization` method should be:
~~~solidity
nonce = keccak256(abi.encodePacked(
    challenge.id,
    challenge.realm,
    challenge.request.externalId // 👈 add externalId
))
~~~

Challenge binding for `permit2`:
~~~solidity
struct PaymentWitness {
    bytes32 challengeHash;
    // 👈 remove the externalId
}

challengeHash = keccak256(abi.encodePacked(
    challenge.id,
    challenge.realm,
    challenge.request.externalId // 👈 move externalId to here
))
~~~

### 3. Remove `challenge.realm` in PaymentWitness
If we can remove `externalId` in Permit2 payment witness struct, then we can also remove `challenge.realm`.
- Same reason as in point 1: `challenge.id` is computed using `base64url(HMAC-SHA256(secretKey, realm👈)|method|intent|serialize(request|expires|digest|opaque))` ([Core](https://github.com/tempoxyz/mpp-specs/blob/ddeb95f28e90085f49ad0413c441b08e0bcfe4a8/specs/core/draft-httpauth-payment-00.md?plain=1#L311-L326))
- `challenge.id` is calculated using `challenge.realm` and already has binding. No need to put it in PaymentWitness again.

### Additional context

_No response_

chopmob-cloud:
Closes #287.

@Xeift is right: `externalId` in the Permit2 `PaymentWitness` is redundant.

The witness commits to `challengeHash = keccak256(challenge.id, challenge.realm)`, and `challenge.id` is an HMAC whose slot 3 is the **entire `request` object**, JCS-serialized per RFC 8785 (core `draft-httpauth-payment-00`). Since `externalId` is a member of `request`, it is already bound to `challenge.id` (and therefore to `challengeHash` and the signature) whenever present. The separate witness field binds nothing new.

Removing it is also a small hardening: the old text carried a `PaymentWitness.externalId == ""` absent-case rule that duplicated, at the witness layer, the absent-optional-field convention the core spec already resolves deterministically at HMAC slot 3 via JCS. Two separately-specified conventions for the same thing is a needless cross-implementation divergence risk.

No regression to the `type="hash"` mitigation ("Requiring unique `externalId` values per challenge") — that operates at the `challenge.id` layer, not the witness, so distinct `externalId` still yields distinct `challengeHash`.

**Changes (one file):**
- Drop `string externalId;` from the `PaymentWitness` struct and from the EIP-712 witness type string.
- Reword the Witness Data section to state binding flows through `challengeHash` -> `challenge.id` (which already commits to every request field, incl. `externalId`).
- Update the `type="permit2"` bullet accordingly.

`externalId` intentionally remains as the request field, the receipt echo, and the hash-mode mitigation.

**Validation:** builds clean through `kramdown-rfc` + `xml2rfc` (HTML/TXT); `rfclint --no-rng --no-spell` exit 0; `lint_frontmatter.py` clean. DCO signed-off.

### DIFF
--- a/specs/methods/evm/draft-evm-charge-00.md
+++ b/specs/methods/evm/draft-evm-charge-00.md
@@ -151,7 +151,7 @@ This specification defines four credential types:
   - The server naturally sponsors gas (fee payer)
   - Split payments are atomic via batch transfers
   - No nonce management burden on the client
-  - `externalId` is cryptographically bound via witness data
+  - `externalId` is cryptographically bound via the challenge hash
 
 - **`type="authorization"`**: The client signs an
   off-chain EIP-3009 {{EIP-3009}} `transferWithAuthorization`
@@ -450,10 +450,12 @@ in array order.
 
 The Permit2 witness mechanism provides cryptographic
 binding between the payment authorization and the
-challenge. When `externalId` is present in the challenge
-request, the client MUST include it in the EIP-712
-witness struct. When `externalId` is absent from the challenge request,
-`PaymentWitness.externalId` is the empty string (`""`). The server MUST
+challenge. The witness commits to `challengeHash`, which
+is derived from `challenge.id` and `challenge.realm`.
+Because `challenge.id` is an HMAC over the serialized
+challenge request (see {{I-D.httpauth-payment}}), every
+request field, including `externalId` when present, is
+already bound through `challengeHash`. The server MUST
 verify the witness matches before submitting the transaction.
 
 The witness type is defined as:
@@ -461,7 +463,6 @@ The witness type is defined as:
 ~~~solidity
 struct PaymentWitness {
     bytes32 challengeHash;
-    string externalId;
 }
 ~~~
 
@@ -480,7 +481,7 @@ against a different challenge, even if the payment
 parameters are identical.
 
 The witness type string for EIP-712 is:
-`"PaymentWitness witness)PaymentWitness(bytes32 challengeHash, string externalId)TokenPermissions(address token,uint256 amount)"`
+`"PaymentWitness witness)PaymentWitness(bytes32 challengeHash)TokenPermissions(address token,uint256 amount)"`
 
 This specification defines that witness schema directly for
 challenge binding. Implementations MUST use the exact type


brendanjryan:
even if redundant, changing this interface would be breaking. I'm preferential to _not_ do this in the short term

Issue #264, opened 2026-05-19
EVM Charge: gas sponsoring for permit2 approve tx


Xeift:
### Describe the feature

I've spent some time reading the EVM Charge spec. In the Permit2 Payload section, it is mentioned:
https://github.com/tempoxyz/mpp-specs/blob/b3695145e03f95fa661eef174a6274184b8047af/specs/methods/evm/draft-evm-charge-00.md?plain=1#L397-L407

So to my understanding, the `permit2` credential type in `evm` requires the client to have signed an approval for the Permit2 contract by themselves in advance, and currently does not support gas sponsoring for the approve tx.

This means the client needs to prepare native coin to pay the gas if they use a traditional EOA, which leads to a bad UX.

I acknowledge that Privy already supports EIP-7702 + Paymaster, but based on the [doc](https://docs.privy.io/wallets/gas-and-asset-management/gas/setup#:~:text=Certain%20flows%20that%20require%20on%2Dchain%20ECDSA%20signature%20verification%20such%20as%20Permit2%20are%20not%20supported), I believe it does not support Permit2.

x402 v2 has introduced the `erc20ApprovalGasSponsoring` extension. When enabled:

1. The client signs an approve EIP-1559 raw tx, and instead of broadcasting it, sends this signed tx to the server.
2. The server, or more accurately, the facilitator, simply transfers ETH to the client's EOA.
3. The server broadcasts the tx from step 1.

Steps 2 and 3 need to be atomic.
On Ethereum, this can be achieved by using an MEV bundle, such as `eth_sendBundle` in Flashbots.
On Base, there seems to be no public method like `eth_sendBundle`. Flashblocks + Private RPC might be a solution, but this does not guarantee atomicity.
On MegaETH, Mini-blocks and `realtime_sendRawTransaction` might be a solution. Similar to Flashblocks, this does not guarantee atomicity.
On Sei, SIP-2 is the exact solution, though it is still in draft. It maintains atomicity.

With `erc20ApprovalGasSponsoring`, x402 clients do not need to have ETH in their EOA and do not need to approve Permit2 in advance. Their SDK handles this automatically. But yes, this increases the complexity of the server/facilitator-side implementation, especially the atomic part. Different chains have different block producers.

x402 uses EIP-3009 (`authorization`) as the main method to settle txs, so agents do not need to have native coin in their EOA; only USDC is required, which is really convenient.

[...]

PR #318, opened 2026-08-09
fix(evm): prefer EIP-3009 authorization

Xeift:
Prefer `authorization` for tokens known to support EIP-3009 when no splits are requested.
Permit2 remains the preferred option otherwise.

Refs #264

### DIFF
--- a/specs/methods/evm/draft-evm-charge-00.md
+++ b/specs/methods/evm/draft-evm-charge-00.md
@@ -82,9 +82,9 @@ method in the Payment HTTP Authentication Scheme
 exchange one-time ERC-20 token transfers on any EVM-compatible
 blockchain.
 
-Four credential types are supported: `type="permit2"`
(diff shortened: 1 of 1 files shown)


chopmob-cloud:
No objection to preferring `type="authorization"` for EIP-3009 tokens when
there are no splits. The line you draw (authorization for simple single
transfers, Permit2 for splits and server-paid fees) is the right one, since
EIP-3009 `transferWithAuthorization` cannot express atomic splits.

One coordination note: #313 also edits this section of
`draft-evm-charge-00.md`. It removes the redundant `externalId` from the
Permit2 `PaymentWitness`, because every request field (including `externalId`
when present) is already bound through `challengeHash`. That change is
orthogonal to the preference change here (witness cleanup vs recommended
type), and I confirmed #313 rebases cleanly on top of this PR with no
conflicts. Happy to sequence #313 after this one merges.

PR #325, opened 2026-08-20
fix(tempo-charge): align the EIP-712 Proof contract with shipped v3

EfeDurmaz16:
The Proof Payload section specifies the v1 EIP-712 contract: domain version `"1"`, message `Proof(string challengeId)`. All three shipping SDKs sign and verify v3, `Proof(address account, string challengeId, string realm)` with domain version `"3"`:

- mpp-go [pkg/tempo/proof.go#L28-L37](https://github.com/tempoxyz/mpp-go/blob/fd01098f563b0bba96ddafd3b55a03ed424a0524/pkg/tempo/proof.go#L28-L37)
- mpp-rs [src/protocol/methods/tempo/proof.rs#L14-L22](https://github.com/tempoxyz/mpp-rs/blob/85c62c59205789d6506ad0c77f543a8680f02612/src/protocol/methods/tempo/proof.rs#L14-L22) (v3 since [#318](https://github.com/tempoxyz/mpp-rs/pull/318))
- mppx [src/tempo/internal/proof.ts#L13-L23](https://github.com/wevm/mppx/blob/a74b20b925d63b4f86962c140808db1858089b88/src/tempo/internal/proof.ts#L13-L23)

The digests diverge, so a server implementing the spec as written recovers a non-matching address and rejects every valid zero-amount proof. For chainId 42431, challengeId `kM9xPqWvT2nJrHsY4aDfEb`, realm `api.example.com`, account `0x1a642f0E3c3aF545E7AcBD38b07251B3990914F1`:

```
spec v1    0xc5c155e34a264f8a8ff363f3e989f402356697303a4e1102d512c25b1b2afb11
shipped v3 0x3860a700a55e02ad3c2dc047e92489feceecbdb0a801d948e1d9f0b61ea9bc3f
```

The v3 value matches the constant mpp-rs pins in `test_signing_hash_matches_mppx_v3_vector`.

This updates the typed data and the Proof Verification steps to the shipped contract. ERC-1271 and on-chain access-key delegation are written as MAY, since mppx and mpp-rs implement delegation as a fallback and mpp-go does not; adjust the strength if you intend otherwise.

Related, not in this PR: stripe/mpp-rb is still on v2 (migration open at stripe/mpp-rb#38) and the mpp-tools tempo-proof vector pins v2.

`make check` and the frontmatter/section-ref linters pass locally. Prepared with AI assistance per CONTRIBUTING; the SDK citations and digests were reproduced independently.

### DIFF
--- a/specs/methods/tempo/draft-tempo-charge-00.md
+++ b/specs/methods/tempo/draft-tempo-charge-00.md
@@ -460,6 +460,23 @@ from the challenge, binding the proof to the issuing origin and
 preventing cross-realm replay of a proof obtained from a different
 server.
 
+### Proof Contract Versions
+
+The domain `version` identifies the proof contract. Three versions have
+shipped:
+
+| Version | `Proof` fields | Description |
+|---------|----------------|-------------|
+| `"1"` | `challengeId` | Initial contract. Binds a proof to a single challenge. |
+| `"2"` | `challengeId`, `realm` | Adds realm binding, so a proof obtained from one server cannot be replayed against another. |
+| `"3"` | `account`, `challengeId`, `realm` | Adds payer binding, so a proof cannot be replayed against a different account, including across an access key authorized for multiple accounts. |
+
+Implementations MUST use version `"3"`. The domain `version` is part of
+the EIP-712 digest, so a signature produced under version `"1"` or
+`"2"` does not verify against the version `"3"` typed data. Servers
+MUST NOT accept proof signatures made under domain version `"1"` or
+`"2"`.
+
 ### Proof Verification
 
 Servers MUST verify proof credentials as follows:
--- a/specs/methods/tempo/draft-tempo-charge-00.md
+++ b/specs/methods/tempo/draft-tempo-charge-00.md
@@ -431,24 +431,34 @@ The typed-data domain and types are:
 {
   "domain": {
     "name": "MPP",
-    "version": "1",
+    "version": "3",
     "chainId": <challenge methodDetails.chainId>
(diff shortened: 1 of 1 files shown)


brendanjryan:
Thank you for this! Instead of overwriting the number, could we outline the three versions for completeness? Happy to do this and carry this fwd

EfeDurmaz16:
Done in ce5853f: added a **Proof Contract Versions** table listing v1 (`challengeId`), v2 (`+realm`, mppx#389), and v3 (`+account`, mppx#532), each row naming the replay its binding closes, with v3 kept as the contract to implement.

Two notes from reconstructing the history:

- It is worth enumerating, because the version string has drifted from the shape twice: mpp-go `95d243d` (2026-05-01) and mpp-rs `51daa18` (2026-05-13) both shipped a changed `Proof` type while the domain `version` stayed `"1"`.
- stripe/mpp-rb still signs v2, so it is non-conformant against this section until stripe/mpp-rb#38 lands.

`make check` passes locally. Happy to adjust the normative strength on v1/v2 if you would rather they stay acceptable.

brendanjryan (approved):

EfeDurmaz16:
@brendanjryan Follow-up in `tempoxyz/mpp-tools`: `conformance/vectors/tempo-proof.json` still pins domain version `"2"`, which this now contradicts. I have the fix ready but PR creation there is restricted, so I filed tempoxyz/mpp-tools#177 for a related fixture bug and can do the same here, or push a branch if you would rather pull it.

PR #326, opened 2026-08-21
spec: add output schemas to payment discovery

epistemedeus:
## Summary

- bump the payment-discovery draft from revision 00 to 01
- recommend that payable operations describe every intended successful response representation with standard OpenAPI response schemas
- scope output claims to the applicable status, media type, and schema branches
- distinguish seller-declared discovery evidence from runtime observation and validation
- add one response-schema example and extend the existing information-disclosure note

## Why

The discovery draft already uses OpenAPI for payment metadata and request construction, but it does not explain how sellers should declare successful response contracts. An agent can therefore discover how to call a paid operation without enough declared evidence to evaluate what a successful response contains.

This revision uses standard OpenAPI semantics. It does not add a new schema dialect or a projection format. A schema for one status and media type cannot stand in for another, and an operation-wide claim is valid only when every declared successful representation entails it.

The runtime 402 challenge remains authoritative for all payment parameters. Output schemas remain advisory seller declarations and do not replace validation of the actual HTTP response.

## Compatibility and safety

- no payment method, intent, amount, currency, challenge, credential, receipt, or settlement behavior changes
- no runtime, SDK, catalog, or ranking code changes
- examples, defaults, unrequired properties, and GraphQL selections outside the OpenAPI document are not treated as requiredness guarantees
- schema interpretation follows the declared OpenAPI version and JSON Schema dialect, with fail-closed guidance when relevant semantics cannot be evaluated

## Validation

- frontmatter lint passed
- external-section-reference lint passed
- kramdown-rfc generated RFC XML
- xml2rfc generated HTML and text artifacts
- rfclint passed with the repository's existing boilerplate warnings
- the exact signed tree received an independent adversarial review for multiple success statuses and media types, object and null semantics, composition, references, array cardinality, payment invariance, and revision structure

The repository's normal Docker build and validation workflow remains the authoritative CI gate.

## AI assistance disclosure

Grok 4.6 xhigh assisted with current-state research and an initial draft. Codex independently reviewed and substantially amended the text against the OpenAPI and JSON Schema specifications. The submitted tree was then independently re-reviewed. I am responsible for the final content.

### DIFF
--- a/specs/extensions/draft-payment-discovery-01.md
+++ b/specs/extensions/draft-payment-discovery-01.md
@@ -1,8 +1,8 @@
 ---
 title: Service Discovery for HTTP Payment Authentication
 abbrev: Payment Discovery
-docname: draft-payment-discovery-00
-version: 00
+docname: draft-payment-discovery-01
+version: "01"
 category: info
 ipr: noModificationTrust200902
 submissiontype: IETF
@@ -83,11 +83,11 @@ the "Payment" HTTP authentication scheme. Services
 publish an OpenAPI document annotated with payment
 extensions that describe pricing, payment methods, and
 intent types. The OpenAPI document serves as the
-canonical machine-readable contract, providing both
-payment metadata and input schemas so that agents can
-discover and invoke endpoints. The runtime 402
-challenge remains authoritative for all payment
-parameters.
+canonical machine-readable contract, providing payment
+metadata and, when declared, input and output schemas
+so that agents can discover and invoke endpoints. The
+runtime 402 challenge remains authoritative for all
+payment parameters.
 
 --- middle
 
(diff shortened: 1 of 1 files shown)


brendanjryan:
Thank you! 

I'm supportive of this for completeness, but would prefer that we indicate in the spec that this is just a re-iteration of the existing OpenAPI spec. 

Broadly I want to keep this extension thin and limited to only additive things, not re-hashing underlying implementations unless

* critical to understanding
* useful for examples

epistemedeus:
Thanks, agreed. I added a short clarification that this section reiterates existing OpenAPI response and schema semantics and defines no new extension or dialect. The remaining text stays scoped to payment-discovery guidance. The one-file amendment is at signed, GitHub-verified head ad01e43f180a86305bd6242a993cddd75ecb4033; the repository frontmatter, external-reference, pytest, RFC XML, HTML, text, PDF, and rfclint checks pass.

brendanjryan:
LGTM -- thanks!

brendanjryan (approved):

PR #328, opened 2026-08-21
Alternate payment credential header

raubrey-stripe:
In step with the proposed change to `mppx`, include a spec change that outlines an 'alternate payment credential header'.

**Proposed new header field in challenge**

```
  HTTP/1.1 402 Payment Required
  WWW-Authenticate: Payment id="pay_abc123",
    realm="api.example.com",
    method="tempo",
    intent="charge",
    // This field says you should not use the default Authorization header, but instead use Payment-Authorization
    header="Payment-Authorization", 
    request="eyJhbW91bnQiOiIxMDAwIn0"
```

If header is not included, the challenge should be included in the Authorization header.

### DIFF
--- a/specs/core/draft-httpauth-payment-00.md
+++ b/specs/core/draft-httpauth-payment-00.md
@@ -1,8 +1,8 @@
 ---
 title: The "Payment" HTTP Authentication Scheme
 abbrev: Payment Auth Scheme
-docname: draft-httpauth-payment-01
-version: "01"
+docname: draft-httpauth-payment-00
+version: 00
 category: std
 ipr: noModificationTrust200902
 submissiontype: IETF
--- a/specs/core/draft-httpauth-payment-00.md
+++ b/specs/core/draft-httpauth-payment-00.md
@@ -103,8 +103,8 @@ Payment Challenge
 Payment Credential
 : An HTTP header field with scheme "Payment" containing payment
   authorization data. The default field name is `Authorization`. A
-  Payment challenge MAY select a different field with the `header`
-  parameter (see {{credential-header}}).
+  Payment challenge MAY select `Payment-Authorization` with the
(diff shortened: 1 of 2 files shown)


specs/core/draft-httpauth-payment-00.md
```diff
@@ -286,6 +322,24 @@ auth-param      = token BWS "=" BWS ( token / quoted-string )
   purpose. This parameter is for display purposes only and MUST NOT be
   relied upon for payment verification (see {{amount-verification}}).
 
+**`header`**: Selects the HTTP field for the Payment credential, as
+  specified in {{credential-header}}. The value MUST be a valid
+  `field-name` as defined by {{RFC9110}}. The default field is
```
brendanjryan: Please constrain this to a closed allowlist of supported, spec-owned fields (initially only `Payment-Authorization`). Accepting any `field-name` can collide with fields such as `Host`, `Content-Length`, or `Proxy-Authorization`; a “not currently registered” rule is unstable as the registry changes.

specs/core/draft-httpauth-payment-00.md
```diff
@@ -820,7 +959,7 @@ while the actual `request` payload requests a different amount.
 
 ## Credential Storage
 
-Implementations MUST treat `Authorization: Payment` headers and
+Implementations MUST treat any header carrying a Payment credential and
```
brendanjryan, line 965: Please also require `Cache-Control: private` (or `no-store`) on every response to a request carrying a Payment credential outside `Authorization`. RFC 9111 §3.5 protects only `Authorization`, and `Payment-Receipt` is optional.

brendanjryan:
This makes sense to me! Left two granular comments but i'm +1 to the high level direction

brendanjryan (approved):

Issue #331, opened 2026-08-22
stellar-charge: push-mode (type=hash) drops expiry check and has no challenge-to-transaction binding, enabling payment reuse


memosr:
The `stellar` charge method's push-mode verification (`type="hash"`) omits checks that its own pull-mode verification performs, and provides no binding between the challenge and the submitted transaction. A server that advertises push mode can be made to accept a payment it did not receive.

## Missing checks in push mode

`specs/methods/stellar/draft-stellar-charge-00.md`, "Push Mode Verification" (~lines 552-570) vs "Pull Mode Verification" (~lines 490-531).

1. **Expiry check dropped.** Both pull flows require: "the current time is before the challenge `expires` auth-param." Push-mode step 1 only requires the `id` to match an outstanding, unsettled challenge; the `expires` clause is absent. An expired challenge is still accepted in push mode.

2. **No payer (`from`) verification.** Push-mode step 4 checks only `to == recipient` and `amount == amount`. The `from` argument is never constrained, so any transaction paying the right amount to the merchant satisfies the check, regardless of who sent it.

3. **No challenge-to-transaction binding.** The spec contains no memo, nonce, or challenge-hash requirement (zero occurrences of memo/nonce/binding/front-running in the document). Nothing ties a specific transaction to a specific challenge.

## Why this is exploitable

The Stellar ledger is public, and `recipient` + `amount` are published in the challenge. On a fixed-price endpoint:

1. Attacker requests the resource -> 402 with `recipient=GMERCHANT...`, `amount=5000000`, `currency=<contract>`.
2. Attacker searches Horizon/RPC for any successful transfer to `GMERCHANT...` of exactly `5000000` of that asset. On a fixed-price resource these are other customers' payments, or the attacker's own earlier payment.
3. Attacker submits `{"type":"hash","hash":"<that tx>"}`.
4. Server runs push-mode steps: id valid, hash not yet consumed by this server, tx SUCCESS, to/amount/currency match -> resource is delivered and a receipt is issued.

The consumed-hash set only prevents a second use of the same hash; it never verifies the first use belongs to the payer of this challenge. Side effect: when the legitimate payer later submits their own hash, it is rejected as already consumed, so their payment is burned.

## Family comparison

[...]

PR #332, opened 2026-08-22
fix(stellar/charge): add expiry, network, and ledger-time checks to push mode verification

memosr:
Push-mode verification (`type="hash"`) in the Stellar charge method omits checks that this same document requires for pull mode, which lets a caller present a transaction that was never made for this challenge.

## What was missing

Comparing "Push Mode Verification" against "Pull Mode Verification" in the same file:

| Check | Pull (sponsored) | Pull (unsponsored) | Push |
|---|---|---|---|
| Current time before `expires` | yes | yes | **no** |
| Network passphrase matches `methodDetails.network` | yes | yes | **no** |
| Any constraint on transaction freshness | `timeBounds` / auth entry expiration | `timeBounds.maxTime` <= `expires` | **no** |

Because the ledger is public and the challenge itself publishes `recipient` and `amount`, an unrelated successful transfer paying that amount to that merchant satisfies every remaining push-mode check. On a fixed-price endpoint such transactions exist: other customers' payments, or the caller's own earlier payment. The consumed-hash set only prevents a second use of the same hash; it does not establish that the first use belongs to this challenge. A side effect is that the legitimate payer's later submission of their own hash is rejected as already consumed.

## What this PR changes

Adds the three checks to push mode, bringing it to parity with pull mode:

1. Challenge expiry check in step 1, matching the wording already used by both pull flows.
2. Network passphrase check, matching pull mode step 4.
3. Ledger close time must fall within the challenge validity window, with transactions confirmed before challenge issuance rejected.

No wire format or client behavior changes; these are server-side verification steps only. The remaining steps are renumbered.

## Deliberately not included

Issue #331 also proposes a challenge-bound transaction memo, which would give push mode the same explicit binding that hedera and tempo have. That does change client behavior, so it is left for discussion in the issue rather than bundled here.

Refs #331

### DIFF
--- a/specs/methods/stellar/draft-stellar-charge-00.md
+++ b/specs/methods/stellar/draft-stellar-charge-00.md
@@ -555,19 +555,30 @@ For push mode credentials (`type="hash"`), the server MUST fetch the
 transaction via Stellar RPC `getTransaction` {{STELLAR-RPC}} and verify:
 
 1. The challenge `id` matches an outstanding, unsettled challenge issued
-   by this server.
+   by this server, and the current time is before the challenge `expires`
+   auth-param.
 
 2. The transaction hash has not been previously consumed (see
    {{replay-protection}}).
 
 3. The transaction exists and has status `SUCCESS`.
 
-4. The transaction contains exactly one `invokeHostFunction` operation
+4. The transaction's network passphrase MUST correspond to
+   `methodDetails.network`.
+
+5. The transaction's ledger close time MUST fall within the challenge
+   validity window. Servers MUST reject transactions that were confirmed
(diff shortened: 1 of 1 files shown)


memosr:
Disclosure per CONTRIBUTING: this change was prepared with AI assistance. The push/pull verification gap was identified by AI-assisted review of the spec, and I verified each cited requirement against the current text on main before filing. I take responsibility for the correctness of the change.

memosr:
Verified locally with `make check` (Docker): all specs build to XML/HTML/TXT/PDF and rfclint reports that all specs passed validation. The rfclint duplicate-word and table-width warnings are pre-existing across the repo and unrelated to this change.

PR #338, opened 2026-08-30
spec: add authorization evidence extension draft

saneGuy:
## Summary

Adds `specs/extensions/draft-payment-authorization-evidence-00.md`, an Informational extension for carrying **authorization evidence** through the Payment flow: verifiable evidence that a principal authorized the presenting agent to make this payment, checked before settlement.

A Payment Credential proves a payment is funded; when the payer is an autonomous agent, it does not prove the agent was permitted to spend. The identity docs already name this boundary ("An attestation doesn't authorize a payment or prove the end user's identity. Apply your own policy after verifying the agent") — this extension gives that policy check an interoperable wire shape instead of leaving it implementation-defined.

## Design

- **Two Challenge parameters** (`evidence`, `evidence-formats`) and **one request header** (`Payment-Authorization-Evidence`). No new round trips; no changes to Credential semantics, settlement, or any core field. Non-implementing clients remain fully conformant per the core's custom-parameter and unknown-field rules.
- **Evidence formats are pluggable** through a "Specification Required" registry, mirroring how the core treats payment methods: this document defines the envelope, six verification properties (challenge binding, presenter binding, amount, payee, temporal validity, principal authenticity), and fail-closed processing — formats define payload schema and trust model.
- **Fail closed per house convention**: verification failure returns `402` with a fresh Challenge; error details SHOULD use RFC 9457 problem details, and evidence failures are reported separately from payment failures.
- Scope explicitly excludes Payment Intents; this is orthogonal to any authorize-and-capture intent.

## Checks

- `make lint`: frontmatter checks pass.
- `make check`: renders clean (HTML/TXT/XML/PDF); the duplicate-word warnings in the output are pre-existing across the corpus (same count with and without this file).
- No hardcoded external section references.

## Disclosure

Per CONTRIBUTING.md: this draft was written with significant AI assistance (Claude and OpenAI Codex, including adversarial review passes); I directed the design, reviewed the generated content for RFC compliance, and take responsibility for the content.

### DIFF
--- a/specs/extensions/draft-payment-authorization-evidence-00.md
+++ b/specs/extensions/draft-payment-authorization-evidence-00.md
@@ -1,0 +1,424 @@
+---
+title: Authorization Evidence Extension for HTTP Payment Authentication
+abbrev: Payment Authorization Evidence
+docname: draft-payment-authorization-evidence-00
+version: "00"
+category: info
+ipr: noModificationTrust200902
+submissiontype: IETF
+consensus: true
+
+author:
+  - name: Viswanadha Pratap Kondoju
+    ins: V. Kondoju
+    email: kondojuviswanadha@gmail.com
+    org: ZKProva Inc.
+
+normative:
+  RFC2119:
+  RFC4648:
+  RFC8126:
+  RFC8174:
+  RFC8259:
+  RFC9110:
+  RFC9457:
+  I-D.httpauth-payment:
+    title: "The 'Payment' HTTP Authentication Scheme"
+    target: https://datatracker.ietf.org/doc/draft-ryan-httpauth-payment/
+    author:
+      - name: Brendan Ryan
+    date: 2026-01
+---
+
+--- abstract
+
+This document defines the "authorization evidence" extension for the
+Payment HTTP Authentication Scheme. It specifies one request header
+field, `Payment-Authorization-Evidence`, and two challenge parameters,
+`evidence` and `evidence-formats`, that let a client present
+verifiable evidence that a principal authorized the paying agent to
+make this payment, alongside the Payment Credential and without
+modifying the core flow.
+
+--- middle
+
+# Introduction
+
+A Payment Credential proves that a payment is funded and well-formed
+under a payment method. When the client is an autonomous agent acting
+for a principal, the Credential does not prove a second fact the
+server may care about: that the principal authorized this agent to
+make this payment — at this amount, to this payee, at this time.
+
+This extension carries that second fact as **authorization evidence**:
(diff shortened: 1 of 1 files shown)


brendanjryan:
Do you have a concrete use-case or evidence format structure? I think this is a good idea, but do not want to merge protocol changes or extensions without concrete need

saneGuy:
Yes. A concrete use case is an agent buying API access under an operator-signed mandate allowing requests under $100, for a specified merchant audience, until expiry. Our published [@bolyra/mpp](https://www.npmjs.com/package/%40bolyra/mpp) implements this through `bolyraGate`. `npx @bolyra/mpp demo` allows a $25 request, denies a $500 request before invoking the payment method, and verifies signed decision receipts. The demo uses real mandate verification with a stub payment method; it settles no funds.

The existing presentation is a versioned `bvp/1` JSON bundle containing `agent`, `binding`, and `sig`. The signed binding contains `agent_name`, `project_key` (merchant audience), `program`, `model`, `capabilities`, and `expiry`. [EVC §4](https://github.com/bolyra/bolyra/blob/main/spec/external-verifier-contract-v1.md) specifies the binding bytes and signature procedure, with [conformance vectors](https://github.com/bolyra/bolyra/blob/main/spec/test-vectors.json). An MPP format profile would still need to specify challenge and presenter binding.

Today, this integration carries evidence in `X-Bolyra-Authorization` by application convention. This PR proposes shared advertisement and carriage through `evidence`, `evidence-formats`, and `Payment-Authorization-Evidence: <format> <payload>`, plus a format registry. The current draft also specifies verification requirements and fail-closed processing. The registry starts empty.

There is a separately authored spending-mandate proposal in whawk46's [x402#3220](https://github.com/x402-foundation/x402/pull/3220). My authorization-evidence extension, [x402#3376](https://github.com/x402-foundation/x402/pull/3376), includes payment-binding inputs for §19-style verification at `7df9973`. Both remain proposals.

If useful, I can trim this PR to the envelope and registry only, leaving format specifications separate.

brendanjryan:
Thanks! Do you have a live integration using this which requires this to be
up streamed? Generally we want to avoid upstreaming anything which is not
widely used as upstream SDKs can easily
Implement this functionality (like in your case)

saneGuy:
No live third-party integration today. Our own gate runs fine out of tree, so by your bar this does not need to be upstream yet, and I think that bar is right. Closing this for now; I will come back if independent adoption makes the case concrete. If a particular kind of evidence would make it worth revisiting, I am glad to know what you would want to see. Thanks for the direct read.

PR #346, opened 2026-09-05
Add XRPL payment method with charge and session intents

krkmu:
## Summary

Adds the `xrpl` payment method for the XRP Ledger, as two documents under
`specs/methods/xrpl/`.

**`draft-xrpl-charge-00`** settles a single Payment transaction, in XRP, an
issued currency, or a Multi-Purpose Token. Two modes: pull, where the client
returns a signed transaction blob and the server submits it, and push, where
the client submits and returns the hash. Registers the `xrpl` method.

**`draft-xrpl-session-00`** covers off-ledger payment channels. The client
signs a cumulative claim per request, the server verifies it locally against
the channel's public key with no round trip to the ledger, and redeems the
last claim when it closes. Registers the `session` intent for `xrpl`. The
method itself is registered by the charge document.

Per the AI tools policy in CONTRIBUTING.md: this draft was written with AI
assistance.

### DIFF
--- a/specs/methods/xrpl/draft-xrpl-charge-00.md
+++ b/specs/methods/xrpl/draft-xrpl-charge-00.md
@@ -1,0 +1,765 @@
+---
+title: XRP Ledger Charge Intent for HTTP Payment Authentication
+abbrev: XRPL Charge
+docname: draft-xrpl-charge-00
+version: 00
+category: info
+ipr: noModificationTrust200902
+submissiontype: independent
+consensus: false
+
+author:
+  - name: Maxime Dienger
+    ins: M. Dienger
+    email: maximed@ripple.com
+    org: RippleX
+
+normative:
+  RFC2119:
+  RFC3339:
+  RFC8174:
+  RFC8259:
+  RFC9457:
+  I-D.payment-intent-charge:
+    title: "'charge' Intent for HTTP Payment Authentication"
+    target: https://datatracker.ietf.org/doc/draft-payment-intent-charge/
+    author:
+      - name: Jake Moxey
+      - name: Brendan Ryan
+      - name: Tom Meagher
+    date: 2026
+  I-D.httpauth-payment:
+    title: "The 'Payment' HTTP Authentication Scheme"
+    target: https://datatracker.ietf.org/doc/draft-ryan-httpauth-payment/
+    author:
+      - name: Jake Moxey
+    date: 2026-01
+
+informative:
+  XLS-33:
+    title: "XLS-33: Multi-Purpose Tokens"
+    target: >
+      https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0033-multi-purpose-tokens
+    author:
+      - org: XRP Ledger Standards
+    date: 2024
+  XRPL-PAYMENT:
+    title: "Payment Transaction"
+    target: https://xrpl.org/docs/references/protocol/transactions/types/payment
+    author:
+      - org: XRP Ledger Foundation
+    date: 2026
+  XRPL-PARTIAL:
+    title: "Partial Payments"
+    target: https://xrpl.org/docs/concepts/payment-types/partial-payments
+    author:
+      - org: XRP Ledger Foundation
+    date: 2026
+  XRPL-CURRENCY:
+    title: "Currency Formats"
+    target: https://xrpl.org/docs/references/protocol/data-types/currency-formats
+    author:
+      - org: XRP Ledger Foundation
+    date: 2026
+  XRPL-TRUSTLINES:
+    title: "Trust Lines and Issuing"
+    target: https://xrpl.org/docs/concepts/tokens/fungible-tokens
+    author:
+      - org: XRP Ledger Foundation
+    date: 2026
+  XRPL-RESERVES:
+    title: "Reserves"
+    target: https://xrpl.org/docs/concepts/accounts/reserves
+    author:
+      - org: XRP Ledger Foundation
+    date: 2026
+  XRPL-FINALITY:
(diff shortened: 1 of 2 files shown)


specs/methods/xrpl/draft-xrpl-session-00.md
```diff
@@ -0,0 +1,780 @@
...
+
+1. the channel exists;
+2. its `Destination` is the recipient this server is charging for --
+   a funder can otherwise open a channel to an address of its own
+   choosing and receive service against claims this server can never
+   redeem;
+3. its `PublicKey` is the key the claim signature was verified
+   against;
+4. its `SettleDelay` is at least the server's configured minimum;
+5. the cumulative claimed does not exceed `Amount` less `Balance`;
```
brendanjryan: `Amount` is the total XRP allocated to the channel, while the signed claim amount is cumulative and `Balance` is what has already been delivered. For example, a valid channel with `Amount=1,000,000`, `Balance=500,000`, and a new cumulative claim of `600,000` fails this check because `600,000 > 1,000,000 - 500,000`.

Can we require `Balance < claimed <= Amount`, then calculate the paid increment from `claimed - max(Balance, server high-water mark)`?

specs/methods/xrpl/draft-xrpl-session-00.md
```diff
@@ -0,0 +1,780 @@
...
+| `description` | string | no | display only |
+
+`amount` is the increment for this request, and the cumulative total
+is the server's to state where it can. A challenge that names a
+channel SHOULD report the mark for it, so a client resumes from
+server state rather than its own bookkeeping.
+
+A challenge that names no channel cannot: there is nothing to look
+the mark up by, and reporting zero is all it can do. A client MUST
+therefore track the highest cumulative it has signed, per network
```
brendanjryan, line 277: This correctly notes that the same channel ID can occur on multiple networks, but the later cache, high-water mark, and finalized-channel requirements key state only by channel ID. That aliases unrelated channels across networks.

Can we require `(network, channelId)` for every server-side key as well? We may also need to make `network` required, since the tuple cannot be formed when it is absent.

specs/methods/xrpl/draft-xrpl-session-00.md
```diff
@@ -0,0 +1,780 @@
...
+shared between them makes the second network sign above what it was
+asked for.
+
+`channelId` is empty when the server names no channel, for either
+of two reasons. On an open-action challenge the channel does not
+exist yet: its ID cannot be read until the creating transaction is
+validated. Otherwise, a server accepting callers it has not met has
+no channel to name, because a client learns its channel ID from its
+own `PaymentChannelCreate`. In both cases the client supplies the
+channel, and a credential payload MUST always carry a full 64-hex
```
brendanjryan: This says every credential payload carries a full `channelId`, but the `open` payload below does not include one. The `close` action also has no JSON schema or field table, and the receipt table defines only Voucher and Open behavior.

Can we define the complete discriminated union and receipt requirements for all three actions? Either `open` needs a `channelId`, or this MUST needs to be scoped to the voucher and close variants.

specs/methods/xrpl/draft-xrpl-charge-00.md
```diff
@@ -0,0 +1,785 @@
...
+## Challenge
+
+~~~ json
+{
+  "method": "xrpl",
+  "intent": "charge",
+  "amount": "1000000",
+  "currency": "XRP",
+  "recipient": "rhewi79quXUDwcqjkpj4bXuw3cuHYC9fwv",
+  "expires": "2026-08-21T10:32:00Z",
```
brendanjryan: The core spec defines `method`, `intent`, and `expires` as `WWW-Authenticate` parameters, with the JCS request object base64url-encoded in `request`. This example instead flattens auth parameters and request fields into one JSON object and omits `id`, `realm`, and `request`.

Can we show the actual `WWW-Authenticate` header followed by the decoded request object? The session challenge example has the same issue.

brendanjryan:
Left some high level comments!

krkmu:
Four findings, fixed in the latest commit. Two were the spec misdescribing the implementation; two reached the SDK.

**Channel State bound.** `Amount` less `Balance` double-counts, as your example shows. Now `Balance < claimed <= Amount`, with the delivered increment as `claimed - Balance` and your `max(Balance, mark)` case. The implementation had the upper bound and was missing the lower one, so it now also refuses a cumulative at or below `Balance`. Monotonicity did not cover it: that compares against the server's own mark, and `Balance` runs ahead of the mark whenever a claim is redeemed outside the exchange.

**State keys.** The asymmetry was mine, required of the client and not of the server. There is now a normative State Keys section over the high-water mark, cached metadata and the finalized record, plus hex canonicalisation, all of which the implementation already did. `network` as `MUST` is implemented as a verification requirement rather than a schema one: the request schema validates what a route supplies, and the server fills `network` in its request hook, so requiring it in the schema would make every route repeat what the server already knows. A challenge omitting it, or naming a network the server does not settle on, is refused.

**The open payload.** `open` carries no `channelId` and cannot, since the channel does not exist until its transaction validates. The MUST is now scoped to `voucher` and `close`, `close` gets its JSON and field table, and the receipt table covers all three actions with `txHash` conditional on `close`. Spec only: the payload union already had this shape.

**Challenge examples.** Both flattened the header parameters into the request object. Replaced with the real `WWW-Authenticate` header and the decoded `request`.

Unrelated: an appendix heading was renamed, it collided with a section of the same name.

Thanks for the prompt review.

brendanjryan (approved):
LGTM

PR #348, opened 2026-09-07
spec(extensions): draft-payment-compliance-00 — payer compliance screening before settlement

OceanAlt666:
## What

A companion extension for the Payment scheme that standardises the three touchpoints the core already leaves open for "payment verified, but policy denies access" (§4.2):

1. `compliance` Challenge parameter (§9.3 additional parameter, ignored by clients that don't know it; not part of the HMAC binding) — the server advertises that the payer in `source` will be screened, against which policy, by whom.
2. `403 application/problem+json` shape for a valid-but-denied payment: `decision`, `reason_code`, `evidence_uri`, `retry`, `challenge_id`. No new `WWW-Authenticate` on 403 — re-paying won't change the outcome.
3. `Payment-Compliance` response header next to `Payment-Receipt`: a compact, challenge-bound screening record so receipts can be audited and disputes resolved (the receipt's stated use).

## What it does not do

- No identity requirement on the payer. It screens the pseudonymous on-chain account already in `source` (§11.7 respected).
- No change to core, no change to any method spec, no new binding rules.
- Fail-closed: a provider outage is `UNCERTAIN`, never `PASS`.

## Why now

Resource servers taking machine payments need to screen the paying account (sanctions / issuer freeze / mixer exposure) before settlement and prove afterwards that they did. Today each implementer invents its own 403 body and its own receipt side-channel. This is the same gap the x402 community is closing with the `onBeforeSettle` trust-provider extension (x402-foundation/x402#2299, draft-03 appendix A lists a compliance provider); MPP has the cleaner hooks for it, so the companion is smaller.

## Reference implementation

A framework-agnostic HTTP gate that wraps any `(Request) => Response` resource handler, with a runnable demo against a live screening endpoint:
https://github.com/OceanAlt666/mpp-compliance-adapter (`mpp.mjs`, `mpp-demo.mjs`, MIT).

Happy to rename the parameter / header, move the problem types under a neutral namespace, or split the receipt part out if the maintainers prefer a smaller first extension.

### DIFF
--- a/specs/extensions/draft-payment-compliance-00.md
+++ b/specs/extensions/draft-payment-compliance-00.md
@@ -1,0 +1,199 @@
+---
+title: "Compliance Screening Extension for the Payment HTTP Authentication Scheme"
+abbrev: "Payment Compliance"
+docname: draft-payment-compliance-00
+category: info
+ipr: trust200902
+submissiontype: independent
+consensus: false
+author:
+  - name: OceanAlt
+    org: OceanAlt
+    email: business@oceanalt.com
+---
+
+# Abstract
+
+This extension lets a resource server advertise, perform, and evidence a compliance screening of the payer before it settles a payment made under the Payment HTTP Authentication Scheme [I-D.httpauth-payment]. It defines one optional Challenge parameter (`compliance`), one problem type family for the `403` response the core specification already reserves for "payment verified, but policy denies access", and one optional response header (`Payment-Compliance`) that carries an auditable screening record next to the `Payment-Receipt`.
+
+# Status of This Memo
+
+This document is a companion extension draft submitted for discussion in the `specs/extensions/` directory. It does not modify the core specification. It is not an Internet Standards Track specification.
+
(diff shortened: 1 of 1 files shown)


brendanjryan:
Thank you! We are not interested in enshrining compliance controls into the core spec at this point, as they are highly variable and application dependent. Clients and servers may still use these primitives and we will monitor future asks for this when deciding when to upstream.

OceanAlt666:
Thanks — understood, and the variability point is fair.

We'll keep this as an external companion implementation and continue
building on MPP's existing extensibility points rather than trying to
standardize prematurely.

What we'll focus on now is the evidence: real usage, independent
implementations, and reproducible interoperability cases. Our reference
adapter and interoperability tests remain public for reproducible
experimentation:

https://github.com/OceanAlt/mpp-compliance-adapter

If the same need starts appearing across independent implementations,
we'll bring back concrete interoperability evidence rather than another
proposal.

Appreciate the guidance.

PR #349, opened 2026-09-08
docs(tempo): cite TEMPO-TX-SPEC in the charge method

devorun:
## Summary

`draft-tempo-charge-00` defines the `TEMPO-TX-SPEC` reference (Tempo Transaction Specification) in its front matter but never cites it in the body, so the reference is unused and does not appear in the rendered References section.

The sibling `draft-tempo-session-00` cites `{{TEMPO-TX-SPEC}}` in the same two sentences — the signed-transaction credential (`type` `"transaction"`, type `0x76` serialized as RLP) and the fee-payer placeholder-signing step. This adds the matching citations to the charge method so it links to the transaction spec it relies on, bringing it to citation parity with the session method.

No normative changes; wrapping of the two affected paragraphs is adjusted to fit the inline citations.


brendanjryan:
👍 thank you!

brendanjryan (approved):

PR #350, opened 2026-09-09
docs(titles): use Title Case for the Stripe and Tempo charge intent titles

devorun:
## Summary

The `title:` front matter of the Stripe and Tempo charge intent specs uses a lowercase "charge":

- `draft-stripe-charge-00`: `Stripe charge Intent for HTTP Payment Authentication`
- `draft-tempo-charge-00`: `Tempo charge Intent for HTTP Payment Authentication`

Every other method charge spec uses Title Case (`Card Network Charge Intent…`, `EVM Charge Intent…`, `Solana Charge Intent…`, etc.). Capitalize "Charge" in these two so the rendered document titles are consistent across the suite.


brendanjryan (approved):
👍

PR #353, opened 2026-09-09
fix: derive problem pages from the core specification

decofe:
Define `bad-request` (400), `invalid-payload` (402), `internal-payment-error` (500), and `payment-action-required` (402) in the core specification. These problem URIs are already emitted by mppx but currently have no documentation pages.

Have the existing page generator read the core spec's Error Codes table directly, so adding an error to the spec also generates its page and index entry through the normal deployment. Link core pages to the actual draft revision. Keep the core draft at revision 01.

Validation: 34 Python tests passed, including a regression test that adds a new spec error and verifies its generated page, example, index entry, and revision link. Both lint checks passed, and all 19 generated problem URLs returned HTTP 200 locally. The generation tests run in the existing CI job. Full spec rendering is covered by CI; Docker is unavailable in the local sandbox.

Prepared with AI assistance.

Prompted by: @brendanjryan


brendanjryan (approved):
cc @parvahuja

PR #355, opened 2026-09-11
feat(solana): advertise accepted transaction versions in charge and session

lgalabru:
## Summary

Adds an optional `methodDetails.transactionVersions` field to the Solana charge and session drafts so a server can advertise which Solana transaction message versions it accepts for client-submitted transactions, and clients can build Transaction V1 ([SIMD-0385](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0385-transaction-v1.md)) once the server supports it.

- **Field**: array of the integers `0` and/or `1` (the Wallet Standard `supportedTransactionVersions` vocabulary). Absent means `[0]`. Legacy messages are not supported; servers MUST NOT advertise `"legacy"` and MUST reject a legacy message. Every deployed MPP client builds version 0 today, so nothing changes for existing deployments.
- **Version 1 rules** (new "Transaction Versions" section in each draft): 4096-byte size limit; compute budget in the message `TransactionConfig`, with `computeUnitLimit` and `loadedAccountsDataSizeLimit` required; no Compute Budget instructions; `priorityFee` is a total in lamports, checked against a per-compute-unit cap as `priorityFee * 1e6 <= maxPriceMicroLamports * computeUnitLimit`. Servers MUST NOT advertise `1` until the `enable_tx_v1` gate is active on the network (live on devnet and testnet; mainnet scheduled for epoch 1035).
- **Verification**: the pull-mode charge verifier and the session open/top-up steps check the message version, size and (on version 1) compute configuration before inspecting instructions; address lookup tables stay rejected.
- The operator-built session `distribute` may be a version-1 transaction instead of a version-0 one with a lookup table.

Companion changes: solana-foundation/x402#11 (same field on the SVM x402 schemes), solana-foundation/pay-kit#319 (Rust reference implementation).

## Notes

- Adds an informative `SIMD-0385` reference to both drafts.
- There is no Solana subscription method draft in this repository; the same rules are documented in the pay-kit SDK references for the subscription intent.
- `make check` was not run locally (no Docker here); the external-section-reference lint passes.

### DIFF
--- a/specs/methods/solana/draft-solana-charge-00.md
+++ b/specs/methods/solana/draft-solana-charge-00.md
@@ -73,6 +73,9 @@ informative:
     author:
       - org: Anza
     date: 2026
+  SIMD-0385:
+    title: "SIMD-0385: Transaction V1 Format"
+    target: https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0385-transaction-v1.md
   BASE58:
     title: "Base58 Encoding Scheme"
     target: https://datatracker.ietf.org/doc/html/draft-msporny-base58-03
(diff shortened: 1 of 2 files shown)


brendanjryan (approved):
LGTM

PR #360, opened 2026-09-17
fix(core): insert HMAC header slot before opaque

### LINKED ISSUE #357: HMAC challenge-id header slot: spec appends after opaque; SDKs insert before
The recommended HMAC-SHA256 challenge-id construction for the optional `header` auth-param disagrees between the draft and both shipping SDKs.

## Spec (append after opaque)

[draft-httpauth-payment-01](https://github.com/tempoxyz/mpp-specs/blob/main/specs/core/draft-httpauth-payment-01.md) § Recommended: HMAC-SHA256 Binding still documents seven fixed slots (`realm|method|intent|request|expires|digest|opaque`) and, when `header` is present, **appends** an eighth slot *after* `opaque`:

```
input = "|".join([
    realm, method, intent, request_b64url,
    expires or "", digest or "", opaque_b64url or "",
])
if header is present:
    input = input + "|" + header
```

See the slot table (slot 6 = `opaque`, slot 7 = `header`) and the Python-shaped example immediately below it in that section. Header-less input is unchanged.

This layout was introduced in #328 (commit ccab885 on draft-00) and was not amended when the draft was promoted to -01.

## SDKs (insert immediately before opaque)

Two independent implementations insert `header` **immediately before** `opaque`, keeping opaque as the final optional slot:

- **mppx** `idBindingInput` — wevm/mppx#821, `src/Challenge.ts` (`values.push(credentialHeader)` then `values.push(opaque)`).
- **mpp-rs** `compute_challenge_id_with_header` — tempoxyz/mpp-rs#400, `src/protocol/core/challenge.rs` (push advertised header, then push opaque).

Worked vector (`secret = "test-vector-secret"`, `realm = "api.example.com"`, `method = "tempo"`, `intent = "charge"`, `request = base64url({"amount":"1000000"})`, no expires/digest/opaque):

| Layout | HMAC input | id (base64url) |
|---|---|---|
| header-less (all three agree) | `api.example.com\|tempo\|charge\|{req}\|\|\|` | `X6v1eo7fJ76gAxqY0xN9Jd__4lUyDDYmriryOM-5FO4` |
| SDK (insert before opaque) | `…\|\|\|Payment-Authorization\|` | `S91xi-OFGZPMs-j7GsX0FDpIkmCcZT1P9XyV58WNy_U` |
| spec (append after opaque) | `…\|\|\|\|Payment-Authorization` | different |

A client and server that each follow a different document cannot verify each other's challenges.

## Request

Please amend the draft so the recommended HMAC input matches the SDK layout (insert `header` immediately before `opaque` when advertised). That is the layout already on the wire for mppx and mpp-rs, and Elixir `MPP` is following those SDKs as a deliberate interoperability exception until the spec catches up.

If the spec layout is the intended one instead, the SDKs need a coordinated breaking change; today they agree with each other and not with the draft.

kutluhaneth46:
## Summary
- Fixes #357: the recommended HMAC-SHA256 challenge-id construction now inserts the optional `header` slot immediately before `opaque`, matching mppx (`idBindingInput`) and mpp-rs (`compute_challenge_id_with_header`).
- Header-less input is unchanged: `realm|method|intent|request|expires|digest|opaque`.
- With `header="Payment-Authorization"` and empty expires/digest/opaque, the input ends `…|||Payment-Authorization|` rather than appending after opaque.

## Why
A client following draft-01 and a server following the shipping SDKs (or the reverse) cannot verify each other's challenges. Both SDKs already agree with each other; this amends the draft to that on-wire layout.

## Test plan
- [x] `python3 scripts/lint_frontmatter.py` passes
- [ ] Confirm CI `make check` / draft build on the PR
- [ ] Spot-check against the worked vector in #357 (`S91xi-OFGZPMs-j7GsX0FDpIkmCcZT1P9XyV58WNy_U` for the SDK layout)


brendanjryan:
Superseded by #362, which keeps the change spec-only and includes precise slot terminology, embedded vectors, and corrected interoperability rationale.

